PR #22, opened 2025-07-01
Analytics service refactor before it gets out of hand

jterry64:


api/app/query.py
```diff
@@ -0,0 +1,55 @@
+from typing import List, Tuple
+
+
+def create_gadm_dist_query(
```
gtempus, line 4: I think a follow up activity would be to see how this works using the [builder pattern](https://refactoring.guru/design-patterns/builder)

api/app/main.py
```diff
@@ -418,90 +394,37 @@ async def get_analytics_result(resource_id: str):
...
+        geojson = await get_geojson(aoi)
 
         if metadata_content["intersections"]:
             intersection = metadata_content["intersections"][0]
         else:
             intersection = None
 
-        alerts_df = await _analyze(geojson, aoi, intersection)
-        
+        alerts_df = await zonal_statistics(geojson, aoi, intersection)
```
gtempus, line 426: We're gonna move to doing this async and right after the `post` aren't we?

test/unit/api/test_dist_analytics.py
```diff
@@ -0,0 +1,127 @@
...
+def test_create_gadm_adm2_dist_query_no_intersection():
+    query = create_gadm_dist_query(["IDN", "24", "9"], [])
+    
+    assert query == strip_extra_whitespace("""
+        SELECT country, region, subregion, alert_date, alert_confidence AS confidence, SUM(count) AS value
+        FROM 's3://gfw-data-lake/umd_glad_dist_alerts/parquet/gadm_dist_alerts.parquet'
+        WHERE country = 'IDN' AND region = 24 AND subregion = 9
+        GROUP BY country, region, subregion, alert_date, alert_confidence
+        ORDER BY country, region, subregion, alert_date, alert_confidence
+    """)
```
gtempus, line 15: I think you could break these up into tests that focus on `SELECT`, `FROM`, `WHERE`, etc.
It would be easier to see what part was broken or needed to change in the future.

jterry64: So just do the same operation 4 times, but assert on a different part of the statement?

test/unit/api/test_dist_analytics.py
```diff
@@ -0,0 +1,127 @@
...
+async def test_get_geojson_from_data_api():
+    async def send_request_to_data_api_test(url, params):
+        return {
+            "data": [
+                {
+                    "gfw_geojson": "{\"type\":\"MultiPolygon\",\"coordinates\":[[[[147.082461217,-37.914061475],[147.079322906,-37.914295051],[147.079382556,-37.913043591],[147.079792868,-37.910992492],[147.07764676,-37.91072299],[147.078204151,-37.907936585],[147.07838676,-37.907876491],[147.082461217,-37.914061475]]]]}"
+                }
+            ],
+            "status": "success"
+        }
```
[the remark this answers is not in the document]
jterry64: When is a good cutoff for using a pytest fixture if it's only in one test? Or at the very least, where do you put pytest fixtures only used in one or two places instead of just dumping it in conftest?

api/app/main.py
jterry64: I might quickly do this as the final refactor for this PR

jterry64:
@gtempus thinking I changed most things you wanted, going to cut myself off here. Lmk if you think this is ready to pass to you after standup!

gtempus:
Yes, let's get this into main and I'll continue from there, @jterry64 . 🤝

PR #24, opened 2025-07-07
Switch to non-clipped versions of datasets

danscales:
Switch to non-clipped versions of datasets
 - Use the non-clipped versions of datasets and do xr.align on the fly.
 - Moved a bunch of duplicated constants to pipelines/globals.py.
 - Made some names consistent (singular rather than plural)

Still need to create a non-clipped zarr of natural lands.


gtempus (approved):
Thanks for cleaning this up, @danscales! 💪 
Also, i learned about `xarray align`!

PR #25, opened 2025-07-07
PZB-169: Refactor to Use Background Task for Zonal Stats and Improve FastAPI Organization

gtempus:
Check [Miro for a dependency diagram](https://miro.com/app/board/uXjVIwcBcEQ=/?moveToWidget=3458764634284701409&cot=14) and how I'm thinking about adding the Natural Lands dataset next.

# Aim of this PR
To start to tease apart the various boundaries of the api app itself and get it into a more common fastAPI-looking form.
There are still several things to address. But I wanted to stop here and get these changes into `main`

## Pull request checklist

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build related changes

The app was basically in one file.

Issue Number: PZB-169

## What is the new behavior?
Now it's more of a fastAPI structure with models, routers, etc.

- [ ] Yes
- [x] No


api/app/routers/land_change/dist_alerts/__init__.py
```diff
@@ -0,0 +1,66 @@
+from .dist_alerts import create
+
+create.__doc__ = """
```
jterry64, line 3: ok, so docs are supposed to go in the init?

api/app/routers/land_change/dist_alerts/dist_alerts.py
```diff
@@ -0,0 +1,346 @@
...
+        "from_attributes": True,
+        "validate_by_name": True,
+    }
+
+
+class DistAlertsAnalyticsResponse(Response):
+    data: DistAlertsAnalytics
+
+
+async def do_analytics(file_path):
```
jterry64, line 241: so we'll still need to put this in a background task next sprint?

api/app/routers/land_change/dist_alerts/dist_alerts.py
```diff
@@ -0,0 +1,346 @@
...
+
+class DataMartResourceLink(StrictBaseModel):
+    link: str
+
+
+class DataMartResourceLinkResponse(Response):
+    data: DataMartResourceLink
+
+
+class AreaOfInterest(StrictBaseModel):
```
jterry64, line 48: Is the best place for all of this still in the same file as the routes?

test/integration/api/test_dist_analytics.py
```diff
@@ -1,10 +1,100 @@
...
+import pytest
+
 import pandas as pd
 from fastapi.testclient import TestClient
 
 from api.app.main import app
 
 client = TestClient(app)
 
+class TestDistAnalyticsWithNoPreviousResult:
```
jterry64, line 12: Ah I like this. So basically you can set up a test per each assertion but have each test be doing the same initial requests?

jterry64 (approved):
This looks good to me with some follow up tasks! I think next sprint we'll need to move things properly to background tasks, and also think how we want to structure adding additional datasets so it doesn't become a gigantic conditional

api/app/routers/land_change/dist_alerts/dist_alerts.py
gtempus: Currently, it runs in a [fastapi BackgroundTask](https://fastapi.tiangolo.com/tutorial/background-tasks/). You can see where it's being used in the [path operation function](https://github.com/wri/project-zeno-data-infra/pull/25/files#diff-61d3d763d429935208b5b0c2abd954414f7a633882c482acff82876cc0c6278dR200) above. As a result, the server sends the response immediately and works on the analytics without blocking.

Obviously, this isn't a sustainable solution. So, yes, we'll design a more robust solution in the next sprint.

test/integration/api/test_dist_analytics.py
gtempus: Yes, this speeds things up significantly with one **very** important caveat. The tests that rely on the setup can't rely on each other. As long as we have test independence, all is well. For each of these tests, there is no further mutation of the payload or application state. We just read different parts of the response. So, it's okay for only one action to happen instead of one for each test.

test/test_dist_create_zarr.py
```diff
@@ -66,18 +65,19 @@ def test_create_zarr_new_file(mock_open_dataset, mock_s3_exists, mock_dataset):
     mock_open_dataset.return_value = mock_dataset
 
     version = "v20250102"
-    cog_uri = f"s3://test-bucket/umd_glad_dist_alerts/{version}/raster/epsg-4326/cog/default.tif"
+    cog_uri = f"s3://gfw-data-lake/umd_glad_dist_alerts/v20250102/raster/epsg-4326/cog/default.tif"
```
danscales, line 68: Just checking - did you mean to leave in the specific version in the file name, as opposed to {version}, and also did you mean to leave the reference to gfw-data-lake, rather than test-bucket?  Same for line 74 and 79/80.  If you really mean to leave the strings as they are currently, then you don't need the f signifier in front of the string.

api/app/routers/land_change/dist_alerts/dist_alerts.py
```diff
@@ -0,0 +1,122 @@
...
+
+    if metadata_data.exists() and analytics_data.exists():
+        return DataMartResourceLinkResponse(data=link, status=AnalysisStatus.saved)
+
+    if metadata_data.exists():
+        return DataMartResourceLinkResponse(data=link, status=AnalysisStatus.pending)
+
+    payload_dir.mkdir(parents=True, exist_ok=True)
+    metadata_data.write_text(payload_json)
+    background_tasks.add_task(do_analytics, file_path=payload_dir)
```
jterry64, line 63: ah, now I see! I think it was loading the diff for me before

api/app/routers/land_change/dist_alerts/dist_alerts.py
```diff
@@ -0,0 +1,122 @@
...
+    # Convert model to JSON with sorted keys
+    payload_dict = data.model_dump()
+
+    # Convert to JSON string with sorted keys
+    payload_json = json.dumps(payload_dict, sort_keys=True)
+
+    # Generate deterministic UUID from payload
+    resource_id = str(uuid.uuid5(uuid.NAMESPACE_DNS, payload_json))
+
+    # Store payload in /tmp directory
```
jterry64, line 46: I suppose this is also a part we might want to think about this sprint - we should save to S3 or a permanent file store!

test/integration/api/test_dist_analytics.py
```diff
@@ -1,75 +0,0 @@
-import pandas as pd
```
jterry64, line 1: @gtempus did you mean to delete this?

api/test/integration/test_dist_analytics.py
```diff
@@ -0,0 +1,168 @@
...
+        }
+    )
+
+    actual_df = pd.DataFrame(data["result"])
+    print(actual_df)
+
+    pd.testing.assert_frame_equal(expected_df, actual_df, check_like=True)
+
+
+def test_kba_dist_analytics_no_intersection():
```
jterry64, line 139: I know you're still working on it, fyi this is the only test that's failing for me. I saw two issues, one is that with the check for metadata I'll post above, and otherwise it gets stuck in pending for the OTF

api/app/routers/land_change/dist_alerts/dist_alerts.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+
+    # Construct file path
+    file_path = PAYLOAD_STORE_DIR / resource_id
+    analytics_metadata = file_path / "metadata.json"
+    analytics_data = file_path / "data.json"
+    metadata_content = None
+    alerts_dict = None
+
+    if analytics_metadata.exists() and analytics_data.exists():
```
jterry64, line 93: I think here you need to still set metadata_content is just analytics_metadata.exists() is true so you can return a pending resource below

gtempus: oh, good catch! Yeah I think I see what you're talking about, @jterry64

api/app/routers/land_change/dist_alerts/dist_alerts.py
gtempus: Think about for sure. But, i don't think we should try to implement it yet. I want to still refactor these routers.

api/app/models/land_change/dist_alerts.py
```diff
@@ -0,0 +1,75 @@
...
+        examples=["2023", "2023-12-31"],
+    )
+    intersections: List[Literal["driver", "natural_lands"]] = Field(
+        ..., min_length=0, max_length=1, description="List of intersection types"
+    )
+
+
+class DistAlertsAnalytics(StrictBaseModel):
+    result: Optional[dict] = {  # column oriented for loading into a dataframe
+        "__dtypes__": {
```
solomon-negusse, line 49: Flagging this for future task - the results here are still in raw format (e.g., confidence and alert_date in `int`)

solomon-negusse (commented):
Nice restructuring @gtempus - i left a couple of comments to note them for future iteration.

PR #26, opened 2025-07-09
Add terraform for service

jterry64:
Add initial terraform for API service to spin up ECS cluster on a load balancer using existing subnets. This spins up the service correctly, but some additional changes are needed in the application code to run properly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,10 +20,9 @@ RUN curl -LsSf https://github.com/astral-sh/uv/releases/download/${UV_VERSION}/u
 RUN useradd -m -s /bin/bash appuser
 RUN mkdir -p /app
 RUN chown appuser:appuser /app
-COPY --chown=appuser . /app
 
 USER appuser
-
+COPY --chown=appuser pyproject.toml uv.lock /app/
 WORKDIR /app
 
 # Create a virtual environment with uv inside the container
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -32,5 +31,10 @@ RUN uv venv ${VENV_DIR} --python ${PYTHON_VERSION} --seed --system-site-packages
 ## Verify GDAL and core Python package installation
 RUN . ${VENV_DIR}/bin/activate \
     && uv sync --locked --no-install-project --no-dev
+ENV PATH=${VENV_DIR}/bin:${USR_LOCAL_BIN}:${PATH}
+ENV VIRTUAL_ENV=/app/.venv
+
+COPY --chown=appuser . /app
```
danscales, line 37: Maybe move 37 up and use it to replace 25 (which is just copying two of the files, but then you copy all of them)?  But I'm also fine if you want to leave these as two separate operations.

terraform/main.tf
```diff
@@ -0,0 +1,158 @@
...
+          essential = true
+          portMappings = [
+            {
+                name          = "api"
+                containerPort = 8000
+                hostPort      = 8000
+                protocol      = "tcp"
+            }
+          ]
+          image     = "public.ecr.aws/b7u8b0a6/analytics:latest"
```
danscales, line 50: Is this a TODO item for later to see if you are now able to go back and use a private image from our private ECR repo?

terraform/main.tf
```diff
@@ -0,0 +1,158 @@
...
+      load_balancer = {
+        service = {
+          target_group_arn = module.alb.target_groups["ex_ecs"].arn
+          container_name   = "api"
+          container_port   = 8000
+        }
+      }
+    
+      enable_cloudwatch_logging = true
+      subnet_ids = ["subnet-0f1544432f2a769d2", "subnet-06be7fcbfc68758ff", "subnet-04591b309ac62bf35"] #["subnet-093dc828845e30d17", "subnet-06a71eea1358f008f", "subnet-061f3f293ed2f3f5e"]  #
```
danscales, line 65: I'm not in any way advocating to fix this now, given how hard it was for you to get this working, but is this a TODO item for later to try to not hardwire the subnets and security group ids?

danscales (approved):
Thanks for working so hard to get this working!

terraform/main.tf
jterry64: Oh true, I can try that now

PR #29, opened 2025-07-14
Get KBA integration test working, use full paths in module imports, fix linting issues

jterry64:
A small PR to get the tests and linting working. In order to run tests from root directory in app folder, using full path for imports instead of relative.

### DIFF
--- a/api/test/integration/test_dist_analytics.py
+++ b/api/test/integration/test_dist_analytics.py
@@ -2,12 +2,11 @@ import json
 import os
 from pathlib import Path
 
-import pytest
-
 import pandas as pd
+import pytest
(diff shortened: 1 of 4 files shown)


api/test/integration/test_dist_analytics.py
```diff
@@ -183,7 +182,7 @@ def setup_before_each(self):
 
     def test_returns_saved_status(self):
         resource = self.test_request.json()
-        actual_df = pd.DataFrame(resource["data"]["result"])
+        # actual_df = pd.DataFrame(resource["data"]["result"])
```
gtempus, line 185: I think you can just delete this. Good catch! 👀

PR #31, opened 2025-07-15
Pzb 259/natural lands classes

solomon-negusse:
Use the correct SBTN natural lands class names and add natural/non-natural category

### DIFF
--- a/api/app/analysis/dist_alerts/analysis.py
+++ b/api/app/analysis/dist_alerts/analysis.py
@@ -1,4 +1,6 @@
 import json
+import logging
+import traceback
 import os
 
 import duckdb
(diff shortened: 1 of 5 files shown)


api/app/analysis/dist_alerts/analysis.py
```diff
@@ -92,7 +93,10 @@ async def zonal_statistics(geojson, aoi, intersection=None):
         alerts_df[aoi["type"]] = aoi["id"]
 
     if intersection == "natural_lands":
-        alerts_df.natural_land_class = alerts_df.natural_land_class.apply(
+        alerts_df["natural_lands_category"] = alerts_df.natural_lands_class.apply(
+            lambda x: "natural" if 1 < x < 12 else "non-natural"
```
danscales, line 97: Didn't know you could do '1 < x < 12' in python!

pipelines/disturbance/gadm_dist_alerts_by_natural_lands.py
```diff
@@ -21,44 +29,280 @@ def gadm_dist_alerts_by_natural_lands(zarr_uri: str, version: str, overwrite: bo
...
-        440, 442, 446, 450, 454, 458, 462, 466, 470, 474, 478, 480, 484, 492, 496,
-        498, 499, 500, 504, 508, 512, 516, 520, 524, 528, 531, 533, 534, 535, 540,
-        548, 554, 558, 562, 566, 570, 574, 578, 580, 581, 583, 584, 585, 586, 591,
-        598, 600, 604, 608, 612, 616, 620, 624, 626, 630, 634, 638, 642, 643, 646,
-        652, 654, 659, 660, 662, 663, 666, 670, 674, 678, 682, 686, 688, 690, 694,
-        702, 703, 704, 705, 706, 710, 716, 724, 728, 729, 732, 740, 744, 748, 752,
-        756, 760, 762, 764, 768, 772, 776, 780, 784, 788, 792, 795, 796, 798, 800,
-        804, 807, 818, 826, 831, 832, 833, 834, 840, 850, 854, 858, 860, 862, 876,
-        882, 887, 894
+        0,
```
danscales, line 46: I just used np.arange(894) everywhere for the ISO values in the restructuring, so this big array and reformatting will go away.

jterry64 (approved):

solomon-negusse:
> Looks good. I'm happy to take another look after you merge with my restructuring change.

Would appreciate it if you can give it one more look now after resolving conflicts @danscales

api/app/analysis/dist_alerts/analysis.py
```diff
@@ -93,7 +94,10 @@ async def zonal_statistics(geojson, aoi, intersection=None):
...
+        )
+        alerts_df["natural_lands_class"] = alerts_df.natural_lands_class.apply(
             lambda x: NATURAL_LANDS_CLASSES.get(x, "Unclassified")
```
danscales, line 101: Just want to make sure I understand.  alerts_df["natural_lands_class"] is the same as alerts_df.natural_lands_class, because alerts_df is a DataFrame?  It's a little confusing that you're using both in the same line, but no need to change if you don't want.

pipelines/disturbance/gadm_dist_alerts_by_natural_lands.py
```diff
@@ -5,35 +5,47 @@
...
             np.arange(731, 1590),  # dates values
-            [1, 2, 3],             # confidence values
+            [1, 2, 3],  # confidence values
         )
         if groups is None
         else groups
     )
 
-    contextual_column_name = 'natural_lands'
+    contextual_column_name = "natural_lands"
```
danscales, line 48: All of the changes in this file are actually just reformatting by your editor (I assume?).  No actual changes.  So you might want to revert this file, to reduce conflicts with your sub-flow PR, or you can keep it either way.

(By the way, many of these seem like pretty unnecessary reformatting, maybe there's an option in your editor to turn most of these off?

danscales (approved):
Looks good, you can just decide if you want to revert the changes to gadm_dist_alerts_by_natural_lands.py (its only changes are editor reformatting).

solomon-negusse:
> Looks good, you can just decide if you want to revert the changes to gadm_dist_alerts_by_natural_lands.py (its only changes are editor reformatting).

Thanks @danscales, merging this as-is now. For the formatting issue, I'm using the standard formatting style that we use everywhere. The best solution is to implement pre-commit hook soon so we don't push unformatted code to begin with.

api/app/analysis/dist_alerts/analysis.py
solomon-negusse: Yes it's the same. I recall pandas used to throw warnings when using dot notation when adding a new column (the first part) but dot notation looks cleaner when accessing columns.

PR #33, opened 2025-07-15
PZB-183 Generate validation data and pandera models

manukala6:
## Pull request checklist

- [ ] Make sure you are requesting to pull a topic/feature/bugfix branch (right side). Don't request your master!
- [ ] Make sure you are making a pull request against the develop branch (left side). Also you should start your branch off our develop.
- [ ] Check the commit's or even all commits' message styles matches our requested structure.
- [ ] Check your code additions will fail neither code linting checks nor unit test.

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


pipelines/disturbance/validate_zonal_statistics.py
```diff
@@ -1,6 +1,108 @@
...
-    # check global stats identical zarr/tile stats
-    # check diff threshold for AOIs
-    # double check against alternative method
-    pass
+    
+    logger = get_run_logger()
+
+    # load local results
+    validation_df = pd.read_csv("../notebooks/validation_stats.csv")
+    logger.info("Loaded validation stats.")
```
jterry64, line 76: @manukala6 it seems like you're creating the stats in a notebook and then checking it here. But I think the intention was to actually run the alternative stats method here, and then compare it to the parquet.

pipelines/disturbance/validate_zonal_statistics.py
```diff
@@ -1,6 +1,195 @@
...
+class NaturalLandsZonalStats(pa.DataFrameModel):
+    countries: Series[str] = pa.Field(isin=isos)
+    regions: Series[int] = pa.Field()
+    subregions: Series[int] = pa.Field()
+    natural_lands: Series[str] = pa.Field(isin=sbtn_natural_lands_classes)
+    area: Series[float] = pa.Field(ge=0)
+
+def generate_validation_statistics(version: str) -> pd.DataFrame:
+    """Generate zonal statistics for the admin area AOI."""
+    gdf = gpd.read_file("../../test/validation_statistics/br_rn.json") # State of Rio Grande do Norte, Brazil
```
danscales, line 78: I believe this code is most likely to be running (whether locally or in a docker) in the top-level directory of the repo, so I would get rid of the '../..' part of the path.

pipelines/disturbance/validate_zonal_statistics.py
```diff
@@ -1,6 +1,195 @@
...
-    # double check against alternative method
-    pass
+    
+    logger = get_run_logger()
+
+    # load local results
+    version = get_latest_version()
+    logger.info(f"Generating validation stats for version {version}.")
+    validation_df = generate_validation_statistics(version)
+    validation_df.to_csv(f"../../test/validation_statistics/dist_{version}_validation.csv", index=False)
```
danscales, line 162: Same here, I would get rid of the '../..'.  And then maybe comment out this line, since we would only want to execute this line in a debugging situation.

pipelines/disturbance/validate_zonal_statistics.py
```diff
@@ -1,6 +1,195 @@
...
     """Validate Zarr to confirm there's no issues with the input transformation."""
-    # check global stats identical zarr/tile stats
-    # check diff threshold for AOIs
-    # double check against alternative method
-    pass
+    
+    logger = get_run_logger()
+
+    # load local results
+    version = get_latest_version()
```
danscales, line 159: **You are missing the required argument to get_latest_version(). Change to:

get_latest_version("umd_glad_dist_alerts")

pyproject.toml
```diff
@@ -15,6 +15,7 @@ dependencies = [
     "numbagg",
     "numpy",
     "pandas",
+    "pandera[pandas]",
```
danscales, line 18: Your validation code is also using 'geopandas', so it has to be listed in pyproject.toml.  We should decide in the huddle when we want to run this validate function.  If during a normal run, then it should be in the dependencies.  If only as test, then it can be added to the 'development' group.

danscales (commented):
Looks mostly good now, but I commented on a several more things to fix.  And as I mention, we should decide in the huddle exactly when this validation test should be run (as so how it should be hooked up into the main dist_flow.py).

The validation will definitely fail even when everything is fixed, since we seem to have some problems with the parquet file we compute.  One new problem I noticed (and added in PZB-298) is that our parquet file has a lot of entries where country, region, or subregion are zero.

pipelines/disturbance/validate_zonal_statistics.py
```diff
@@ -1,6 +1,195 @@
...
+    alert_date: Series[int] = pa.Field(ge=731, le=1640) # julian date between 2023-01-01 to latest version
+    confidence: Series[int] = pa.Field(ge=2, le=3) # low confidence, high confidence
+    value: Series[int]
+    
+    class Config:
+        coerce = True
+        strict = True
+        name = "ZonalStatsSchema"
+        ordered = True
+        unique_columns = ["country", "region", "subregion", "alert_date", "confidence"]
```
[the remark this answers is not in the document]
manukala6: Yes, I was able to run the script separately from the notebook after adding pandas as a dependency. 
Screenshot 2025-07-25 at 8 52 46 AM

danscales (approved):
OK, I think you have to merge with Daniel's changes to pyproject.toml and uv.lock.  And then the one remaining thing is to add a call to validate_zonal_statistics() from the main flow code in gadm_dist_alerts.py.  I think you can just add a small shim function there annotated with @task, that calls to validate_zonal_statistics() with the correct argument and call that shim function at the end of dist_alerts_count().

PR #36, opened 2025-07-17
Separate prefect flows

solomon-negusse:
This isolates each pipeline (dist, dist by natural lands, etc.) into separate prefect [subflows](https://docs-2.prefect.io/latest/concepts/flows/#composing-flows) for better observability into each pipeline, parallelization, run control, etc.

This prefect UI view shows task stream (top) for DIST alerts count and DIST alerts count by natural lands subflows (with tasks in the latter expanded) and subflows tab that shows duration and status of each subflow (bottom)
prefect_task_stream
prefect_subflows

### DIFF
--- a/test/test_dist_create_zarr.py
+++ b/test/test_dist_create_zarr.py
@@ -57,7 +57,7 @@ def test_decode_alert_data():
     assert list(result.y.values) == [0, 1]
 
 
-@patch("pipelines.disturbance.create_zarr.s3_object_exists")
+@patch("pipelines.disturbance.create_zarr.s3_uri_exists")
 @patch("pipelines.disturbance.create_zarr.xr.open_dataset")
 def test_create_zarr_new_file(mock_open_dataset, mock_s3_exists, mock_dataset):
     """Test creating a new zarr file when it doesn't exist."""
@@ -75,8 +75,7 @@ def test_create_zarr_new_file(mock_open_dataset, mock_s3_exists, mock_dataset):
         assert result == expected_uri
 
         mock_s3_exists.assert_called_once_with(
-            "gfw-data-lake",
-            "umd_glad_dist_alerts/v20250102/raster/epsg-4326/zarr/umd_glad_dist_alerts.zarr/zarr.json",
+            "s3://gfw-data-lake/umd_glad_dist_alerts/v20250102/raster/epsg-4326/zarr/umd_glad_dist_alerts.zarr/zarr.json",
         )
 
         mock_open_dataset.assert_called_once_with(cog_uri, chunks="auto")


pipelines/disturbance/prefect_flows/gadm_dist_alerts_by_drivers.py
```diff
@@ -0,0 +1,58 @@
...
+from prefect import flow, task
+
+
+from pipelines.disturbance import stages
+from ...globals import DATA_LAKE_BUCKET
+
+
+@task
+def load_data(dist_zarr_uri: str):
+    contextual_uri = f"s3://{DATA_LAKE_BUCKET}/umd_glad_dist_alerts_driver/zarr/umd_dist_alerts_drivers.zarr"
```
danscales, line 16: I would prefer that you put contextual uri in the main function dist_alerts_by_drivers_count, and same with expected_groups below in setup_compute(), and then pass them to these task functions.  That way, all the distinct information for the analysis is in the main analysis function (like what I had before).  And, all these task functions become identical, so maybe they can be made common.  Otherwise, you actually increased the duplication from what I had before.

pipelines/disturbance/prefect_flows/gadm_dist_alerts_by_drivers.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+@task
+def postprocess_result(result: xr.Dataset):
+    return stages.create_result_dataframe(result)
+
+
+@task
+def save_result(result_df: pd.DataFrame, dist_version: str):
+    return stages.save_results(result_df, dist_version, "dist_alerts_by_drivers")
```
danscales, line 46: Same here, can you pass the result_filename "dist_alerts_by_drivers" in from the main function, so this function is common?

pipelines/disturbance/prefect_flows/gadm_dist_alerts_by_drivers.py
```diff
@@ -0,0 +1,58 @@
...
+        np.arange(854),  # subregion codes
+        np.arange(5),  # driver categories
+        np.arange(731, 1590),  # dates values
+        [1, 2, 3],  # confidence values
+    )
+    return stages.setup_compute(datasets, expected_groups, "driver")
+
+
+@task
+def compute_zonal_stat(dataset: xr.DataArray, groupbys: Tuple, expected_groups: Tuple):
```
danscales, line 35: At the very least, this function is repeated in multiple analyses with the exact same definition, can't this common function be moved to prefect_common.py and shared?  Again, not sure why the duplication.  Same for postprocess_result().  And also the other @task functions here once they take all their inputs from the main analysis function.

danscales (commented):
I feel like this restructuring increased the amount of duplicated code from my restructuring.  So, I gave some review comments about getting back to less duplicated code, and trying to keep all the differences about each analysis in the main analysis function.

The sub-flows seems fine, as far as I can tell.

gtempus (changes requested):
I do see what @danscales is saying. It's also really good to see all the duplication so that we know exactly what to generalize. So, I do also appreciate @solomon-negusse going for the pragmatic approach, first.

Given that the duplication is really easy to see now, we value the child flows, and it would be great to have child flows that are _experts_ in their own context, would something like this work?

`pipelines/disturbance/prefect_flows/dist_alerts.py`
```python
from typing import Optional, Tuple
import xarray as xr
import pandas as pd

from prefect import flow, task

from pipelines.disturbance import stages
from ...globals import DATA_LAKE_BUCKET


@task
def load_data(dist_zarr_uri: str, contextual_path: Optional[str] = None):
    contextual_uri = None
    if contextual_path is not None:
        contextual_uri = f"s3://{DATA_LAKE_BUCKET}/{contextual_path}"

    return stages.load_data(dist_zarr_uri, contextual_uri)


@task
def setup_compute(datasets: Tuple[xr.Dataset, ...], expected_groups, contextual_name: Optional[str] = None) -> Tuple:
    return stages.setup_compute(datasets, expected_groups, contextual_name)


@task
def compute_zonal_stat(dataset: xr.DataArray, groupbys: Tuple, expected_groups: Tuple):
    return stages.compute(dataset, groupbys, expected_groups)


@task
def postprocess_result(result: xr.Dataset):
    return stages.create_result_dataframe(result)


@task
def save_result(result_df: pd.DataFrame, dist_version: str, result_filename: str):
    return stages.save_results(result_df, dist_version, result_filename)


@flow(name="DIST Alerts Like a Boss? 🤷‍♂️")
def dist_alerts_child_flow(
        dist_zarr_uri: str,
        dist_version: str,
        contextual_path: Optional[str],
        expected_groups: Tuple,
        contextual_name: Optional[str],
        result_filename: str,
):
    datasets = load_data(dist_zarr_uri, contextual_path)
    compute_input = setup_compute(datasets, expected_groups, contextual_name)
    result_dataset = compute_zonal_stat(*compute_input)

    result_df = postprocess_result(result_dataset)
    result_uri = save_result(result_df, dist_version, result_filename)

    return result_uri
```

----

And then _expert_ parent flows like:
`pipelines/disturbance/prefect_flows/gadm_dist_alerts.py`
```python
import numpy as np
from prefect import flow, task
from .dist_alerts import dist_alerts_child_flow


@flow(name="DIST alerts count")
def dist_alerts_count(dist_zarr_uri: str, dist_version: str):
    return dist_alerts_child_flow(
        dist_zarr_uri=dist_zarr_uri,
        dist_version=dist_version,
        contextual_path=None,
        expected_groups=(
            np.arange(894),  # country ISO codes
            np.arange(86),  # region codes
            np.arange(854),  # subregion codes
            np.arange(731, 1590),  # dates values
            [1, 2, 3],  # confidence values
        ),
        contextual_name=None,
        result_filename="dist_alerts",
    )
```

`pipelines/disturbance/prefect_flows/gadm_dist_alerts_by_drivers.py`
```python
import numpy as np
from prefect import flow
from .dist_alerts import dist_alerts_child_flow


@flow(name="DIST alerts count by drivers")
def dist_alerts_by_drivers_count(dist_zarr_uri: str, dist_version: str):
    return dist_alerts_child_flow(
        dist_zarr_uri=dist_zarr_uri,
        dist_version=dist_version,
        contextual_path="umd_glad_dist_alerts_driver/zarr/umd_dist_alerts_drivers.zarr",
        expected_groups=(
            np.arange(894),  # country ISO codes
            np.arange(86),  # region codes
            np.arange(854),  # subregion codes
            np.arange(5),  # driver categories
            np.arange(731, 1590),  # dates values
            [1, 2, 3],  # confidence values
        ),
        contextual_name="driver",
        result_filename="dist_alerts_by_drivers",
    )
```

`pipelines/disturbance/prefect_flows/gadm_dist_alerts_by_natural_lands.py`
```python
import numpy as np
from prefect import flow
from .dist_alerts import dist_alerts_child_flow


@flow(name="DIST alerts count by natural lands")
def dist_alerts_by_natural_lands_count(dist_zarr_uri: str, dist_version: str):
    return dist_alerts_child_flow(
        dist_zarr_uri=dist_zarr_uri,
        dist_version=dist_version,
        contextual_path="sbtn_natural_lands/zarr/sbtn_natural_lands_all_classes.zarr",
        expected_groups=(
            np.arange(894),  # country ISO codes
            np.arange(86),  # region codes
            np.arange(854),  # subregion codes
            np.arange(22),  # natural lands categories
            np.arange(731, 1590),  # dates values
            [1, 2, 3],  # confidence values
        ),
        contextual_name="natural_lands",
        result_filename="dist_alerts_by_natural_lands",
    )
```

I like this. It has expert child flows and a generalized dist_alert template. 
Hopefully, it will work as expected?

The tests all pass which also is nice since they are decoupled from the prefect pipeline dependencies.

solomon-negusse:
Updated in 699cd38a710cde940b4b728151801f176c48b976 @danscales @gtempus

gtempus (approved):
Nice! I get it now. `with_options` is a partial application that lets you add the specifics later. Each child flow is even more independent than my template. Therefore, we can adjust different child flows as needed in the future.

I am very disappointed that you didn't add any emojis to the flow names. ☹️ 😆

danscales (commented):
I am taking over your PR.  I am making a bunch of fixes, as indicated in my other comments.

Note that the analyses tests got a lot slower (maybe 10 seconds each now, rather than 0.5 seconds).  I'm just guessing, but I think it may be because you are no longer mocking the expected groups, so we have the full value range (?).

PR #38, opened 2025-07-21
Support multiple AOIs

jterry64:
Support multiple AOI requests for both precomputed and OTF AOIs. This required the following changes:
- Store the dask client in the lifespan for usage during requests.
- Make breaking interface change to restrict multiple IDs to one AOI type. Now, instead of accepting multiple `aois`, the API accepts one `aoi` with a type and list of `ids`.
- Refactor to allow for divide process between multiple dask tasks. Currently doing the simplest way: for each `id`, run a separate dask task for both OTF and precalc. For OTF, to avoid spamming Data API with requests, send one request to get all geojsons up front.
- Refactor tests to follow the class-based setup method. In order to test the lifespan required using async requests for the test client.

TODO: for the life of me, I can't shake this notebook change even though I've tried to reset the file multiple times. Will fix it before I merge.


api/app/analysis/common/analysis.py
```diff
@@ -26,30 +27,33 @@ async def get_geojson_from_data_api(aoi, send_request=send_request_to_data_api):
         )
         raise ValueError("Unable to get GeoJSON from Data API.")
 
-    geojson = json.loads(response["data"][0]["gfw_geojson"])
-    return geojson
+    geojsons = [json.loads(data["gfw_geojson"]) for data in response["data"]]
+    return geojsons
```
danscales, line 31: The function name is confusing, since you are now returning a list of geojsons, right?  Can you either rename the function, or maybe just add a header comment explaining that it returns a list?

api/app/analysis/dist_alerts/analysis.py
```diff
@@ -48,7 +50,29 @@
 }
 
 
-async def zonal_statistics(geojson, aoi, intersection=None):
+async def zonal_statistics_on_aois(aois, dask_client, intersections=[]):
+    if intersections:
+        intersection = intersections[0]
```
danscales, line 55: I don't really understand what 'intersections' is, but just checking - is it really right that you only want to look at the first element of intersections, even if it has multiple elements?

api/app/analysis/common/analysis.py
```diff
@@ -26,30 +27,33 @@ async def get_geojson_from_data_api(aoi, send_request=send_request_to_data_api):
...
+        sql = f"select gfw_geojson from data where objectid in {value_list} order by objectid"
     else:
         raise ValueError(f"Unable to retrieve AOI type {aoi['type']} from Data API.")
     return url, {"sql": sql}
 
 
 async def get_geojson(aoi, geojson_from_predfined_aoi=get_geojson_from_data_api):
-    if aoi["type"] == "geojson":
-        geojson = aoi["geojson"]
+    if aoi["type"] == "feature_collection":
```
danscales, line 55: Very minor: rename geojson_from_predfined -> geojson_from_predefined

api/test/integration/test_dist_analytics.py
```diff
@@ -1,59 +1,72 @@
...
-                "aois": [{"type": "admin", "id": "IDN.24.9"}],
-                "start_date": "2024-08-15",
-                "end_date": "2024-08-16",
-                "intersections": [],
-            },
-        )
-
-    def test_post_returns_pending_status(self):
-        resource = self.test_request.json()
+        delete_resource_files("71f40812-2157-5ce2-b654-377e833e5f73")
```
danscales, line 21: Any reason why this ID had to change everywhere from 9c4b... to 71f4.... ?

Is there any requirement on that file name (right format?), or will it have to change again? It would be nice if it could just be "tmp_resource_file".

api/app/models/common/areas_of_interest.py
```diff
@@ -62,32 +62,32 @@ def set_version_default(cls, v):
...
 
 
 class IndigenousAreaOfInterest(AreaOfInterest):
     type: Literal["indigenous_land"] = "indigenous_land"
-    id: str = Field(
+    id: List[str] = Field(
         ...,
         title="Landmark Indigenous lands object ID",
-        examples=["1931", "1918", "43053"],
+        examples=[["1931"], ["1918", "43053"]],
```
dmannarino, line 84: I'm a little confused here. Looking at the examples, shouldn't the type be List[List[str]] ?

api/test/integration/test_dist_analytics.py
dmannarino: I think it's a hash, and I imagine it changing is the result of the input changing subtly (id - >ids, str -> List[str]). But I might be wrong.

api/test/integration/test_dist_analytics.py
```diff
@@ -206,22 +219,199 @@ def test_returns_200_Ok_response_code(self):
         assert response.status_code == 200
 
 
-def test_gadm_dist_analytics_no_intersection():
-    delete_resource_files("9c4b9bb5-0ecd-580d-85c9-d9a112a69b59")
+class TestDistAnalyticsPostWithMultipleAdminAOIs:
+    @pytest_asyncio.fixture(autouse=True)
```
dmannarino, line 223: I see you explicitly naming the setup fixture for the tests below (in order to get the request and client objects). So could we make this not autouse=True? I find lots of autouse fixtures slows things down and can lead to confusing problems, so try not to use them.

api/test/integration/test_dist_analytics.py
```diff
@@ -319,17 +510,16 @@ def write_data_file(dir_path, data):
...
-        data = client.get(
-            f"/v0/land_change/dist_alerts/analytics/{resource_id}"
-        ).json()["data"]
+        resp = await client.get(f"/v0/land_change/dist_alerts/analytics/{resource_id}")
+        data = resp.json()["data"]
         status = data["status"]
+        time.sleep(1)
         attempts += 1
     if attempts >= 10:
         pytest.fail("Resource stuck on 'pending' status")
```
dmannarino, line 525: Oye. This test might take 10 seconds? :(

api/app/analysis/dist_alerts/analysis.py
jterry64: @gtempus while we only support 1 intersection via the API itself, should we just only store 1 intersection, so it's clear across the whole code until we support multiple?

api/test/integration/test_dist_analytics.py
jterry64: Yeah, it's a little fragile because changing the input at all for the test will completely change the hash. Maybe we should just force a specific ID for tests, since it's pretty irrelevant to most tests what the ID actually is?

api/test/integration/test_dist_analytics.py
jterry64: Sure, I actually just copied and pasted this and forgot what autouse actually does...

api/test/integration/test_dist_analytics.py
jterry64: Going to revisit the multiple AOI tests to make sure they don't use real data and can run fast.

api/app/analysis/dist_alerts/analysis.py
gtempus: I would go ahead and make it clear that it's one intersection. We enforce this constraint in the router via the pydantic model. So, remove `intersections` from the codebase to avoid confusion (which is really important).

@danscales, `intersection` is the [term the Zeno team uses](https://github.com/wri/project-zeno-data-infra/blob/1cd9cee0c8e3f0771915707ff6cd05fdcabc4d6b/api/app/routers/land_change/dist_alerts/__init__.py#L20) to describe contextual layers. Since the API is serving them, I agree that we should use their terms to [build the ubiquitous language](https://martinfowler.com/bliki/UbiquitousLanguage.html).

api/app/analysis/dist_alerts/analysis.py
```diff
@@ -111,68 +135,112 @@ async def zonal_statistics(geojson, aoi, intersection=None):
...
+            data = file_path / "data.json"
+            data.write_text(json.dumps(alerts_dict))
     except Exception as e:
         logging.error(
             {
                 "event": "dist_alerts_analytics_processing_failure",
                 "severity": "high",  # Helps with alerting
                 "metadata": metadata_content,
-                "generated_query": query,
+                # "generated_query": query,
```
gtempus, line 243: I'd go ahead and just remove this, @jterry64

api/app/models/common/areas_of_interest.py
```diff
@@ -62,32 +62,32 @@ def set_version_default(cls, v):
...
-        ..., title="Key Biodiversity Area site code", examples=["36", "18", "8111"]
+    ids: List[str] = Field(
+        ..., title="Key Biodiversity Area site codes", examples=[["36"], ["18", "8111"]]
     )
 
 
 class ProtectedAreaOfInterest(AreaOfInterest):
     type: Literal["protected_area"] = "protected_area"
-    id: str = Field(
+    id: List[str] = Field(
```
gtempus, line 72: Shouldn't all of these AOI fields be renamed to `ids`?
```suggestion
    ids: List[str] = Field(
```

api/test/integration/test_dist_analytics.py
gtempus: I think it could be even simpler. When tests run, every artifact should go in some sort of test container/directory. The integration tests should use something like `/tmp/tests/` then the setup hook deletes the entire contents before each run. No need to keep track.

I don't think you have to do this now but it should be simplified one way or another, soon.

api/test/integration/test_dist_analytics.py
gtempus: Adding the `LifespanManager` context definitely made the integration setup messier. I have another idea for the integration tests and how it can be made more sustainable.

api/test/integration/test_dist_analytics.py
gtempus: Just a reminder that:
1. integration tests are slow
2. we shouldn't have many of them
3. focus on unit/collaboration testing and small datasets for quick feedback as @jterry64 mentions

gtempus (changes requested):
@jterry64, I stepped through your commit history on this PR and it looked great! 💪 
With few exceptions, each commit was small and very focused on a specific task. It made it really easy to watch the code evolve. 💯 

Also, your refactoring discipline is starting to show. I saw several instances of _extract method_ happening and tests were failing early in the commit history. Well done! 🏆

api/test/integration/test_dist_analytics.py
jterry64: Will wait on this, I think we're going to change the tests based on Gary's refactoring

api/test/integration/test_dist_analytics.py
jterry64: Will wait on this, I think Gary came up with a good refactoring to avoid slow tests like this

PR #47, opened 2025-07-29
grasslands endpoint and otf support for non-gadm aois

solomon-negusse:


api/app/analysis/grasslands/analysis.py
```diff
@@ -0,0 +1,170 @@
...
+    gadm_id = id.split(".")
+
+    query = create_gadm_grasslands_query(gadm_id, parquet_file)
+    grasslands_df = duckdb.query(query).df()
+
+    grasslands_df["aoi_id"] = id
+    grasslands_df["aoi_type"] = "admin"
+
+    columns_to_drop = ["country"]
+    if len(gadm_id) == 2:
```
danscales, line 96: I think you want this to be >= 2, else you will be leaving the 'region' column around when len(gadm_id) is equal to 3 (probably the normal case).

api/app/analysis/grasslands/analysis.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+
+    table = "gadm_grasslands_areas"
+    local_parquet_file = _download_parquet(table)
+
+    precompute_partial = partial(
+        get_precomputed_statistic_on_gadm_aoi, parquet_file=local_parquet_file
+    )
+    futures = dask_client.map(precompute_partial, aoi["ids"])
+    results = await dask_client.gather(futures)
```
danscales, line 77: I had to get rid of this await, else I got an error message (about can use await on a list of objects).  Once I got rid of it things worked.

danscales: This may have been an artifact of how I was testing by running do_analytics() directly.  The await seems to work fine during the integration tests.

gtempus (changes requested):
Just a few changes requested. Also, thank you for the excellent test efforts and your patience as we get the new hexagonal design to permeate all the endpoints. 🙇 

Also, excellent management of your size and focus of your commits. A majority are focused, short, and sweet. It was really easy to follow your progress! 💯

pipelines/dist_flow.py
```diff
@@ -59,11 +60,11 @@ def dist_alerts_flow(overwrite=False) -> list[str]:
         logger.info(f"Latest dist version: {dist_version}")
         dask_client, _ = create_cluster()
 
-        nl_result = nl_prefect_flow.gadm_natural_lands_area(overwrite=overwrite)
-        result_uris.append(nl_result)
+        # nl_result = nl_prefect_flow.gadm_natural_lands_area(overwrite=overwrite)
+        # result_uris.append(nl_result)
```
danscales, line 64: I assume you will undo most of these changes before check-in ?

danscales (approved):
Looks good to me (after you choose what changes may be left to do from Gary's comments).

gtempus (approved):
Thanks, @solomon-negusse!

PR #49, opened 2025-07-30
PZB-297 Endpoint to serve sbtn_natural_lands table

danscales:
PZB-297 Endpoint to serve sbtn_natural_lands area queries

Does both the precalculated lookup to the parquet and the OTF analysis.

I am following Solomon's changes so that the AOI id includes all GADM components in one string, etc. (And I grabbed one bug fix from his code in app/analysis/common/analysis.py.)

I added in integration tests (which fully pass and exercise the precalculated lookup and the OTF calculation), but not yet unit tests.


api/app/analysis/natural_lands/analysis.py
```diff
@@ -0,0 +1,191 @@
...
+    natural_lands.name = "natural_lands_class"
+    pixel_area = read_zarr_clipped_to_geojson(pixel_area_obj_name, geojson)
+
+    groupby_layers = [natural_lands]
+    expected_groups = [np.arange(1, 22)]
+
+    counts = xarray_reduce(
+        pixel_area,
+        *tuple(groupby_layers),
+        func="count",
```
solomon-negusse, line 78: func  should be "sum"

solomon-negusse: May be worth adding unit tests that would catch this, like this one https://github.com/wri/project-zeno-data-infra/blob/989cb39b5f6430f4a7045742aa6c7cb8663d6e75/api/test/unit/test_grasslands_analysis.py#L115.

api/app/analysis/natural_lands/analysis.py
```diff
@@ -0,0 +1,191 @@
...
+
+    df["natural_lands_class"] = df.natural_lands_class.apply(
+        lambda x: NATURAL_LANDS_CLASSES.get(x, "Unclassified")
+    )
+
+    df = df[df.natural_lands_area > 0]
+    return df
+
+
+async def get_precomputed_statistics(aoi, dask_client):
```
solomon-negusse, line 102: I have refactored this in the grasslands PR taking the download out but that may be updates we address in a refactor

solomon-negusse (changes requested):

api/app/analysis/natural_lands/analysis.py
```diff
@@ -0,0 +1,191 @@
...
+        CREATE OR REPLACE SECRET secret (
+            TYPE s3,
+            PROVIDER credential_chain,
+            CHAIN config
+        );
+    """
+    )
+
+    # table = "gadm_natural_lands_areas"
+    table = "gadm_adm2"
```
dmannarino, line 124: Is it intentional to leave this commented-out?

api/app/models/land_change/natural_lands.py
```diff
@@ -0,0 +1,56 @@
...
+            "country": "str",
+            "region": "int",
+            "subregion": "int",
+            "natural_lands_area": "float32",
+        },
+        "country": ["BRA", "BRA", "BRA"],
+        "region": [1, 1, 1],
+        "subregion": [12, 12, 12],
+        "natural_lands_area": [4851452000, 4084129000, 253008000],
+    }
```
gtempus, line 44: I just posted about this in @solomon-negusse 's PR review. Here's a better way to provide these hints to Swagger:
https://github.com/wri/project-zeno-data-infra/pull/47#discussion_r2248210469

gtempus (changes requested):
Thanks for your patience as we get the app into a more proper structure. I know there have been many volatile, moving parts, @danscales. 🙇 

I do agree with @solomon-negusse that some unit tests (especially for the analysis) would be great and helpful for our future selves.

Also, if you could keep your commits on the smaller side and just squash once the PR is approved, it would provide more insight to your reviewers and encourage small, focused steps during development. I would encourage everyone to create draft PRs as soon as possible and push to them frequently. At the least, it protects your work from getting lost and enables someone to pick up your PR if you need to hand it off.

api/app/analysis/natural_lands/query.py
```diff
@@ -0,0 +1,35 @@
+from typing import Tuple
+
+
+def create_gadm_natural_lands_query(gadm_id: Tuple[str, int, int], table: str) -> str:
```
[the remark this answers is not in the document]
danscales: Even though the 2nd and 3rd elements of the tuple are optional, I think we may get better type checking if we leave as is with Str/int/int.  It don't currently get any mypy errors, even though the 2nd & 3rd elements are optional.

api/app/analysis/natural_lands/analysis.py
danscales: I added a comment in that line 123 is what the eventual name should be, but the current parquet file name is just "gadm_adm2".  I'll be adjusting path and file names in a future PR.

gtempus (approved):
Thanks, @danscales!

PR #57, opened 2025-08-06
fix qc issues 

solomon-negusse:
This addresses alignment issue and DIST alert end date in the past that were both resulting in wrong alert count across all contextual layers


pipelines/prefect_flows/common_stages.py
```diff
@@ -53,15 +53,26 @@ def load_data(
 
     dist_alerts = _load_zarr(dist_zarr_uri)
 
-    country = _load_zarr(country_zarr_uri)
+    # reindex to dist alerts to avoid floating point precision issues
+    # when aligning the datasets
+    # https://github.com/pydata/xarray/issues/2217
+    country = _load_zarr(country_zarr_uri).reindex_like(
```
jterry64, line 59: Should we just move the reindex_like into the _load_zarr function then, or is there a case where you don't want to do this?

danscales (approved):
Thanks for figuring out these issues!

We can discuss in the huddle, but should we always use reindex_like() when using a COG-derived zarr as a basis for alignment?  Do we think this doesn't happen if the zarr is derived from raster tiles?

PR #59, opened 2025-08-06
Grasslands analysis

manukala6:
Create the grasslands parquet file in the pipeline, following the original notebooks/grasslands_areas_gadm_2000-2022.ipynb
See pipelines/grasslands/prefect_flows/grasslands_flow.py and pipelines/grasslands/stages.py.

Using source zarr s3://gfw-data-lake/gfw_grasslands/v1/zarr/natural_grasslands_4kchunk_new.zarr/ from Solomon.

Converted to passing around xr.DataArray in most places, since grasslands operates mostly on DataArrays.  Result of the xarray.reduce() was always an xr.DataArray, declared that.


pipelines/grasslands/prefect_flows/grasslands_flow.py
```diff
@@ -0,0 +1,61 @@
...
+    contextual_column_name = "grasslands"
+    result_uri = "s3://gfw-data-lake/gfw_grasslands/tabular/zonal_stats/gadm/gadm_adm2.parquet"
+    funcname = "sum"
+
+    if not overwrite and s3_uri_exists(result_uri):
+        return result_uri
+
+    expected_groups = (
+        np.arange(894),     # country iso codes
+        np.arange(1, 86),   # region codes
```
solomon-negusse, line 29: We should include 0 in region and subregion to make sure we get correct numbers when aggregating up.

solomon-negusse (changes requested):
I left some comments Dan. And I'd revert the changes to the notebooks completely as it's all exploratory changes.

pipelines/grasslands/prefect_flows/grasslands_flow.py
danscales: OK, done!

pipelines/grasslands/stages.py
```diff
@@ -0,0 +1,66 @@
...
+        grasslands.coords['y'].min().item() - 0.00025, grasslands.coords['y'].max().item()
+    )
+
+    pixel_area: xr.DataArray = _load_zarr(pixel_area_uri).band_data.sel(x=slice(xmin, xmax), y=slice(ymax, ymin)).chunk({'x': 2000, 'y': 2000})
+
+    country: xr.DataArray = _load_zarr(country_zarr_uri).band_data.sel(x=slice(xmin, xmax), y=slice(ymax, ymin)).chunk({'x': 2000, 'y': 2000})
+    region: xr.DataArray = _load_zarr(region_zarr_uri).band_data.sel(x=slice(xmin, xmax), y=slice(ymax, ymin)).chunk({'x': 2000, 'y': 2000})
+    subregion: xr.DataArray = _load_zarr(subregion_zarr_uri).band_data.sel(x=slice(xmin, xmax), y=slice(ymax, ymin)).chunk({'x': 2000, 'y': 2000})
+
+    grasslands.coords['x'] = pixel_area.coords['x']
```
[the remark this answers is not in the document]
danscales: Unfortunately, just changing these two lines to:

grasslands.reindex_like(pixel_area, method="nearest", tolerance=1e-5)

causes thing to fail with "'Object being grouped must be exactly aligned with every array in `by`.'".  So, I think I will leave things as is for now (unless this is not what you meant, and you can show me exactly what you meant.)

solomon-negusse (approved):

pipelines/grasslands/prefect_flows/grasslands_flow.py
danscales: OK, thanks! Not sure why I misread it.  I'll fix before checkin.

PR #60, opened 2025-08-07
Tree cover loss analysis

jterry64:
It ain't pretty, but this is what I could get done today. But even though it's a mess, our repository abstraction at least making it so it has no real dependencies and we can completely test the analysis logic independently! The only thing that seemed awkward is the `compute_engine` part, because in this case, what we need is so some `send_query` type function that can take a dataset/version/query (or something) and send it Data API. And the tests need to be able check the dataset/version/query is correct. So we might need to play around with what works well there.


gtempus (approved):
@jterry64 and I both worked closely on this. I'm happy with this first iteration. 💯 🙌

PR #61, opened 2025-08-07
randomize polygon

solomon-negusse:
The test polygon had a side that aligned exactly with the center of a pixel, leading to inconsistent results across different systems. Due to floating-point precision issues, edge-touching pixels may or may not be included, depending on the underlying platform. I have replaced this with a more realistic/randomized polygon.


gtempus (approved):
I'm glad you solved the mystery, @solomon-negusse! 😅 🔍 🕵️

PR #64, opened 2025-08-11
add support for custom aois feature collection 

solomon-negusse:


solomon-negusse:
> Thank you, @solomon-negusse! I'm sorry that you had to repeat the fix in several places, but, we should have a refactor for that very soon. 🙇

No worries, @gtempus.

PR #65, opened 2025-08-12
PZB-260: Refactor to Common Router and Service

gtempus:
## Pull request checklist

- [ ] Make sure you are requesting to pull a topic/feature/bugfix branch (right side). Don't request your master!
- [ ] Make sure you are making a pull request against the develop branch (left side). Also you should start your branch off our develop.
- [ ] Check the commit's or even all commits' message styles matches our requested structure.
- [ ] Check your code additions will fail neither code linting checks nor unit test.

## Pull request type

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


api/app/routers/common_analytics.py
```diff
@@ -0,0 +1,99 @@
...
+            response.headers["Retry-After"] = "1"
+            message = "Resource is still processing, follow Retry-After header."
+        case AnalysisStatus.saved:
+            message = "Analysis completed successfully."
+        case AnalysisStatus.failed:
+            message = "Analysis failed. Result is not available."
+        case _:
+            message = ""  # Optional default
+
+    return TreeCoverLossAnalyticsResponse(
```
jterry64, line 91: You're returning TCL response in the common function

jterry64 (changes requested):
Found one minor issue. This is very helpful though, thanks! I might tweak how it declares the compute engine/analyzer when swapping to DuckDB. 

One general thing we don't need to change now, but the word `analysis` is really starting to confuse me when reading the code. At the very least, I don't think we should call our code for reading and writing results `analysis_repository`, because it's getting confused with the `analyzer` code to do actually do an analysis. Maybe `resource_repository` or something like that, to distinguish between the data we're storing and the action we're performing to get the data?

api/app/routers/common_analytics.py
gtempus: Woops! Fixed in e4c575142946efa26b2df60235c9e7e392a800cf

PR #70, opened 2025-08-13
Dist alerts area analysis

manukala6:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

GADM Dist Alert analyses results how alert count

Issue Number: [PZB-287](https://gfw.atlassian.net/browse/PZB-287)

## What is the new behavior?

- Switches alert_count to area_ha for GADM dist alerts, GADM dist alerts by natural lands, and GADM dist alerts by driver
- Adds pixel_area fixture and updates tests
- Updates all function names to use `area` instead of `count`

- [x] Yes
- [ ] No

6 tests and 3 function names were changed to reflect alert area instead of alert count

## Other information

[PZB-287]: https://gfw.atlassian.net/browse/PZB-287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pipelines/prefect_flows/common_stages.py
```diff
@@ -68,6 +69,10 @@ def load_data(
         dist_alerts, method="nearest", tolerance=1e-5
     )
     subregion_aligned = xr.align(dist_alerts, subregion, join="left")[1]
+    pixel_area = _load_zarr(pixel_area_uri).reindex_like(
+        dist_alerts, method="nearest", tolerance=1e-5
+    )
+    pixel_area_aligned = xr.align(dist_alerts, pixel_area, join="left")[1]
```
jterry64, line 75: unrelated question for @solomon-negusse since I may have messed this up elsewhere, do you still need to do `align` even after `reindex_like`? Is the `reindex_like` just the snap, and the `align` is the clip?

pipelines/disturbance/stages.py
```diff
@@ -19,17 +19,19 @@ def setup_compute(
     expected_groups: Optional[ExpectedGroupsType],
     contextual_column_name: Optional[str] = None,
 ) -> Tuple:
-    """Setup the arguments for the xrarray reduce on dist alerts"""
-    dist_alerts, country, region, subregion, contextual_layer = datasets
+    """Setup the arguments for the xarray reduce on dist alerts"""
+    dist_alerts, country, region, subregion, pixel_area, contextual_layer = datasets
 
-    mask = dist_alerts.confidence
+    mask = pixel_area.band_data
```
jterry64, line 25: minor naming, but I think when calculating area, this is functioning less as a "mask" than an aggregate layer.

jterry64 (approved):
This looks good to me other than minor style comment, but I'd wait for Dan or Solomon PR since they know this code best right now

pipelines/prefect_flows/common_stages.py
solomon-negusse: Exactly @jterry64, the groupby operation expects all layers to be lined up.

danscales (commented):
As part of this change, I think we would also want you to change the needed query in the webserver for the new parquet files. (@jterry64 would you say that should be in this change, or another change?).   For example in create_gadm_dist_query, you need to at least change "count" to "area_ha".  Similarly, you may also want to change the OTF analysis in dist_alerts/analysis.py:zonal_statistics (which is essentially doing the pipeline on the fly for small areas) to use pixel area as the base layer, and use "sum" rather than "count" for the aggregation.  Maybe talk to Justin on Friday (or I can talk to you again on Monday).

solomon-negusse (changes requested):
@manukala6 This should follow the pattern that's used in natural lands pipeline where the pixel area is the primary layer and natural lands is the contextual layer. For DIST, pixel area will be the primary layer and dist alert confidence and date will be the contextual layer. 

As this stands, natural lands will be broken as it uses the common `load_data` function [primary/base layer](https://github.com/wri/project-zeno-data-infra/blob/main/pipelines/natural_lands/prefect_flows/nl_flow.py#L34)  as you now have added pixel area layer that it has already added as primary layer. That's on us though as we missed to update that function's args and docs when it got pulled out of the disturbance pipeline, which made this confusing. So that function should be:

```
def load_data(
    primary_zarr_uri: str,
    contextual_uri: Optional[str] = None,
) -> Tuple[xr.Dataset, ...]:
    """Load in the primary data Zarr, the GADM zarrs, and possibly a contextual layer zarr"""

    primary_data = _load_zarr(primary_zarr_uri)

    # reindex to primary data to avoid floating point precision issues
    # when aligning the datasets
    # https://github.com/pydata/xarray/issues/2217
    country = _load_zarr(country_zarr_uri).reindex_like(
        primary_data, method="nearest", tolerance=1e-5
    )
    country_aligned = xr.align(primary_data, country, join="left")[1]
    region = _load_zarr(region_zarr_uri).reindex_like(
        primary_data, method="nearest", tolerance=1e-5
    )
    region_aligned = xr.align(primary_data, region, join="left")[1]
    subregion = _load_zarr(subregion_zarr_uri).reindex_like(
        primary_data, method="nearest", tolerance=1e-5
    )
    subregion_aligned = xr.align(primary_data, subregion, join="left")[1]

    if contextual_uri is not None:
        contextual_layer = _load_zarr(contextual_uri).reindex_like(
            primary_data, method="nearest", tolerance=1e-5
        )
        contextual_layer_aligned = xr.align(
            primary_data, contextual_layer, join="left"
        )[1]
    else:
        contextual_layer_aligned = None

    return (
        primary_data,
        country_aligned,
        region_aligned,
        subregion_aligned,
        contextual_layer_aligned,
    )
```
Looping in @danscales for visibility

danscales:
@solomon-negusse Thanks for your comments.  I'm not sure about your described approach, because dist alerts also needs another optional contextual layer of drivers/natural_lands/grasslands, so dist layer can't be the only contextual layer.  So, you would have to have multiple contextual layers.  Maybe it is better to give DIST its own private load function which includes pixel_area, whereas the common load does not?  Maybe you and Maanas can discuss today?

PR #72, opened 2025-08-13
Use parquet and flox for TCL

jterry64:
Switching to use parquet and flox for TCL for consistency and to support all AOI types. Refactoring out how we do parquet and flox analyses to use chain of responsibility and create injectable services for parquet/flox query logic.


api/app/domain/analyzers/data_api_tree_cover_loss_analyzer.py
```diff
@@ -0,0 +1,214 @@
+from abc import abstractmethod
+
+import pandas as pd
+from app.domain.analyzers.analyzer import Analyzer
+from app.domain.models.analysis import Analysis
+from app.models.common.analysis import AnalysisStatus, AnalyticsIn
+from app.models.land_change.tree_cover_loss import TreeCoverLossAnalyticsIn
+
+
+class DataAPIAnalyzer(Analyzer):
```
gtempus, line 10: I don't think we need this Analyzer anymore, right? I can find no usage in the branch.

gtempus (changes requested):
I had one small change request for this PR.

However, i think there are several opportunities to relocate these classes out of the domain since they speak of specific, volatile technologies (extrinsic complexity) like FLOX, ZARR, etc.

What I propose is that we merge the branch in and I'll nudge them in the right place when I adopt this new compute design for Tree Cover Gain. Then you'll see what I think would be better homes for these classes.

PR #73, opened 2025-08-15
PZB-253 Analysis/pipeline for dist with grasslands contextual layer.

danscales:
- Pipeline to generate admin areas for dist with grasslands as contextual layer
 - OTF analysis for dist with grasslands contextual layer
 - Analysis using parquet generated from pipeline for admin areas.
 - Added a couple of integration tests.
 - Some fixes to the pipeline after Maanas' change from Dist counts -> areas.


api/app/analysis/dist_alerts/analysis.py
```diff
@@ -136,13 +152,18 @@ async def zonal_statistics(aoi, geojson, intersection=None):
...
+        alerts_df = alerts_df.drop("year", axis=1).reset_index(drop=True)
 
     alerts_df = alerts_df[alerts_df.value > 0]
     return alerts_df
 
 
-async def get_precomputed_statistics(aoi, intersection, dask_client):
-    if aoi["type"] != "admin" or intersection not in [None, "natural_lands", "driver"]:
+async def get_precomputed_statistics(aoi, intersection: Optional[str], dask_client):
+    if aoi["type"] != "admin" or intersection not in [None, "natural_lands", "driver", "grasslands"]:
```
solomon-negusse, line 166: Minor so you can ignore but I think we can rely on the intersection spec in the DistAnalyticsIn model to handle invalid intersections at the endpoint level or we can use one list in both places.

solomon-negusse (approved):

PR #78, opened 2025-08-21
Get OTF integration test for TCL

jterry64:
A couple issues with OTF I fixed. One thing we may need to discuss is the best way to translate between pixel values we store (e.g. 1-24 for TCL year) and the actual value we use in our queries (2001-2024). For now, I just tried to keep it all in the dataset repository since it's close to the actual data storage.


gtempus (approved):
I approve with no comments because I consider this a continuation of #72 where I shared feedback.
You added tests that caught the issues, so I'm good with approving.

PR #82, opened 2025-08-25
Consistent field names

jterry64:
- Update tree cover loss API to use consistent field names per our discussion
- Update parquet to also use those as field names
- Move parquet to bucket in the same account so we can avoid downloading the whole file

### DIFF
--- a/api/test/integration/test_dist_analytics.py
+++ b/api/test/integration/test_dist_analytics.py
@@ -288,20 +288,23 @@ class TestDistAnalyticsPostWithMultipleAdminAOIs:
 
         expected_df = pd.DataFrame(
             {
-                "country": ["IDN", "IDN", "IDN", "IDN", "BRA"],
-                "region": [24, 24, 14, 14, 1],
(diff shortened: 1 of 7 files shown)


api/test/integration/test_dist_analytics.py
```diff
@@ -288,20 +288,23 @@ async def test_resource_calculate_results(self, setup):
 
         expected_df = pd.DataFrame(
             {
-                "country": ["IDN", "IDN", "IDN", "IDN", "BRA"],
-                "region": [24, 24, 14, 14, 1],
-                "subregion": [9, 9, 13, 13, 1],
                 "aoi_id": ["IDN.24.9", "IDN.24.9", "IDN.14.13", "IDN.14.13", "BRA.1.1"],
                 "aoi_type": ["admin"] * 5,
-                "alert_date": [
+                "disturbance_alerts_date": [
```
danscales, line 293: Just a reminder to undo the changes to this file (since you didn't make corresponding changes elsewhere, but also we agreed not to add the extra long descriptive info for the result fields.)

PR #84, opened 2025-08-26
PZB-289 Dummy endpoint for carbon flux

danscales:
PZB-289 Dummy endpoint for carbon flux

We just return emissions, removals, and net flux (for the period 2000-2024).  There are no start year/end year arguments.

I followed the model of tree_cover_gain.  I didn't add documentation yet, will create it once we have the real analysis.


api/app/models/land_change/carbon_flux.py
```diff
@@ -0,0 +1,91 @@
...
+class CarbonFluxAnalyticsIn(AnalyticsIn):
+    aoi: Annotated[AoiUnion, Field(discriminator="type")] = Field(
+        ...,
+        title="AOI",
+        description="AOI to calculate in.",
+    )
+
+
+class CarbonFluxResult(StrictBaseModel):
+    id: List[str]
```
jterry64, line 34: Based on our agreement with DevSeed, this should be called `aoi_id`, and you should also have an `aoi_type` field that just includes the aoi type as a string

api/app/models/land_change/carbon_flux.py
```diff
@@ -0,0 +1,91 @@
...
+    ProtectedAreaOfInterest,
+    IndigenousAreaOfInterest,
+    CustomAreaOfInterest,
+]
+
+DATE_REGEX = r"^\d{4}$"
+
+
+class CarbonFluxAnalyticsIn(AnalyticsIn):
+    aoi: Annotated[AoiUnion, Field(discriminator="type")] = Field(
```
jterry64, line 26: I think you do need to support canopy cover option here, for thresholds 30, 50 and 75

api/app/models/land_change/carbon_flux.py
```diff
@@ -0,0 +1,91 @@
...
+    data: CarbonFluxAnalytics
+
+    model_config = {
+        "json_schema_extra": {
+            "examples": [
+                {
+                    "data": {
+                        "result": {  # column oriented for loading into a dataframe
+                            "aoi_id": ["BRA.1.12", "BRA.1.12", "BRA.1.12"],
+                            "aoi_type": ["admin", "admin", "admin"],
```
jterry64, line 62: this is correct, just might want to add to the Result model above

api/app/models/land_change/carbon_flux.py
```diff
@@ -0,0 +1,91 @@
...
+                                4050.4061608627,
+                                4045.406160862687,
+                            ],
+                            "carbon_gross_removals_Mg_CO2e": [
+                                4045.406160862687,
+                                4050.4061608627,
+                                4045.406160862687,
+                            ]
+                        },
+                        "metadata": {
```
jterry64, line 79: canopy cover

jterry64 (changes requested):
Set up looks good, just a couple clarification on the interface and results

gtempus (approved):
This looks great from a design standpoint, @danscales! 💯

api/app/models/land_change/carbon_flux.py
danscales: Thanks for catching that! Just a mistake in copying code.

api/app/models/land_change/carbon_flux.py
danscales: Done.

api/app/models/land_change/carbon_flux.py
danscales: Done, and added to the tests as well.

api/app/models/land_change/carbon_flux.py
```diff
@@ -0,0 +1,97 @@
...
+
+AoiUnion = Union[
+    AdminAreaOfInterest,
+    KeyBiodiversityAreaOfInterest,
+    ProtectedAreaOfInterest,
+    IndigenousAreaOfInterest,
+    CustomAreaOfInterest,
+]
+
+AllowedCanopyCover = Literal[30, 50, 75]
```
dmannarino, line 22: Does this work with these as ints instead of strings? I suppose they get coerced?

danscales: I just used the same pattern as appears in api/app/models/land_change/tree_cover_loss.py.  We can discuss at standup if people prefer specifying the choices as numbers or strings.

jterry64 (approved):

PR #85, opened 2025-08-26
Use LCL bucket for parquets instead of GFW data lake

jterry64:


danscales (commented):
Oh, I see now that Justin changed to using a new pixel_area_ha.zarr that he created in the last few days.  So, I think we are good on the area conversion front.  And I also noticed that in a bunch of tests, it looks like he didn't change the expected area numbers, but he actually added a "/ 10000" at the end of the area array list.

PR #132, opened 2025-09-12
Otf dask

solomon-negusse:
This separates out the dask cluster from the API service into a set of dask scheduler and dask cluster manager services. The dask manager is in charge of starting and scaling the cluster. In addition, this:

- adds autoscaling policy to the api service
- increases the api test GHA server size from 2-core to 4-core and it runs twice as fast
- starts the work to create isolated testing environments using terraform workspaces


dmannarino (approved):
Hot diggity! Nice work!

dmannarino (approved):

PR #136, opened 2025-09-15
Correctly handle data out of bounds

jterry64:


gtempus (approved):

api/app/domain/compute_engines/handlers/otf_implementations/flox_otf_handler.py
```diff
@@ -59,7 +59,9 @@ def _handle(aoi, query, dataset_repository, expected_groups_per_dataset):
 
         by = xr.Dataset()
         for ds in query.aggregate.datasets:
-            xarr = dataset_repository.load(ds, geometry=aoi_geometry)
+            xarr = dataset_repository.load(ds, geometry=aoi_geometry).reindex_like(
+                by, method="nearest", tolerance=1e-5
+            )
```
danscales, line 64: The first time through the loop, by is an empty dataset, so the reindex_like operation seems pretty weird.  Are you pretty sure that the reindex_like() operation is a no-op in that case, as opposed to it might actually reduce the indexing of the datasets?  To be safe, do you want to skip the reindex_like the first time through (when 'by' is empty)?

PR #137, opened 2025-09-15
Use shared DuckDB query class for natural lands

jterry64:
- Use DuckDB query service for natural lands
- Make sure query string has correct tuple structure for multiple AOI IDs
- Make sure function is async

### DIFF
--- a/api/test/integration/test_dist_analytics.py
+++ b/api/test/integration/test_dist_analytics.py
@@ -407,42 +407,13 @@ class TestDistAnalyticsPostWithMultipleAdminAOIs:
             ANALYTICS_NAME, analysis_params.thumbprint(), client
         )
 
-        expected_df = pd.DataFrame(
-            {
-                "country": ["IDN", "IDN", "IDN", "IDN", "BRA"],
-                "region": [24, 24, 14, 14, 1],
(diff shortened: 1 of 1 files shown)


api/app/domain/analyzers/natural_lands_analyzer.py
```diff
@@ -77,23 +78,22 @@ async def analyze(self, analysis: Analysis):
...
         await self.analysis_repository.store_analysis(
             natural_lands_analytics_in.thumbprint(), analyzed_analysis
         )
 
-    @staticmethod
-    def analyze_admin_areas(gadm_ids) -> pd.DataFrame:
-        query = f"select natural_lands_class, area_ha, aoi_id from '{admin_results_uri}' where aoi_id in {gadm_ids}"
-        initialize_duckdb()
-        df = duckdb.query(query).df()
+    async def analyze_admin_areas(self, gadm_ids) -> pd.DataFrame:
```
danscales, line 92: The return type here is not  correct anymore, should be a Dict, right?  (Same for the return value of DuckDbPrecalcQueryService.execute()).  But should be fine if you don't want to fix that now.

PR #148, opened 2025-09-19
Fix key error for some params in tree cover gain

dmannarino:
I fixed the key error with TreeCoverGain, with minimal other changes (additional tests, and pre-commit reformatted a few lines).

## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Attempts to query tree cover gain over more than one five-year period fail

Issue Number: N/A

## What is the new behavior?

- One can query tree cover gain for any (single or multiple) five year period(s).

- [ ] Yes
- [x] No

### DIFF
--- a/api/app/domain/analyzers/tree_cover_gain_analyzer.py
+++ b/api/app/domain/analyzers/tree_cover_gain_analyzer.py
@@ -1,6 +1,6 @@
 from typing import List
-import newrelic.agent as nr_agent
 
+import newrelic.agent as nr_agent
 from app.domain.analyzers.analyzer import Analyzer
 from app.domain.compute_engines.compute_engine import ComputeEngine
 from app.domain.models.analysis import Analysis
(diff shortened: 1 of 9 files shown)


jterry64 (approved):
Looks good to me, which line was causing the issue exactly?

dmannarino:
> > Looks good to me, which line was causing the issue exactly?
> > Thanks for asking! I have to admit that I misunderstood the bug the first few times around, and my solutions were heavy-handed (even though I think they generally improved the code). In order to answer I have re-examined it, and found that a far-more minimal fix can be made. I'll PR it shortly and close this one. But to answer your question, it comes from the fact that in the tree cover gain analyzer (specifically in the _build_years helper) we make the filters into a string-encoded tuple of periods (like `"('2000-2005', '2005-2010')"`). Then in zarr_dataset_repository.py we turn in back into a tuple of strings by doing a `ast.literal_eval(value)` on the encoded value. Then we iterate over it the tuple, etc.. The problem is that when there is only one period, _build_years doesn't use a trailing comma for the monotuple (i.e. it outputs `"('2000-2005')"` instead of `"('2000-2005',)"`). But if you want it to be a tuple, Python NEEDS that trailing comma to know it's a tuple, otherwise the parenthesis are superfluous and it evaluates to just a string, `"2000-2005"`. Which zarr_dataset_repository.py then tries to iterate over and gets a key error when it can't find the key `"2"` in the dictionary, the first character in the string `"2000-2005"`. Thank you for challenging me.

So I tried making a few-line modification to _build_years, which went well. Unfortunately when it came time to render the filters in the SQL string those extra commas came out and ruined everything. To avoid doing what I did in this PR I would have to add special handling in the pre-calc SQL query builder, which I really don't want to do. Therefore I think what I have here is a good solution.

danscales (approved):

solomon-negusse (approved):

PR #164, opened 2025-10-06
Dist Versioning

jterry64:


api/app/analysis/dist_alerts/analysis.py
```diff
@@ -240,7 +240,7 @@ def get_precomputed_table(aoi_type: str, intersection: Optional[str]) -> str:
     else:
         raise ValueError(f"No way to calculate aoi type {aoi_type}")
 
-    return f"s3://lcl-analytics/zonal-statistics/{table}.parquet"
+    return f"s3://lcl-analytics/zonal-statistics/dist-alerts/{table}.parquet"
```
danscales, line 243: You should be accessing ..../zonal-statistics/dist-alerts/{latest-finished-version}/{table{.parquet here, right?  Where latest-finished-version is the version you got from the dist-alerts/latest file (so you'll need to pass it in here.

solomon-negusse (commented):
Since this doesn't move the zarr files to zeno account, we have to pass two sets of aws creds. Right now the env variables for the pipeline task runners point to gfw account, so the parquet file writing will fail.  Quickest fix may be to add zeno aws env variables [here](https://github.com/wri/project-zeno-data-infra/blob/main/terraform/prefect.tf#L130) and pass them as `storage_options` to `to_parquet`?

gtempus (approved):
The `api/app` changes look good, @jterry64!

api/app/analysis/common/analysis.py
```diff
@@ -89,6 +89,9 @@ def read_zarr(uri):
     )
 
 
+os.environ["API_KEY"] = "cfb38b37-045a-4157-b96b-aa27c8c700e2"
```
dmannarino: Meant to be in here?

PR #165, opened 2025-10-07
Limit GADM levels in Admin AOI validator

dmannarino:
## Pull request type

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

One can specify an admin ID of specificity greater than level 2 (ex: USA.1.1.1.1.1.1).

Issue Number: N/A

## What is the new behavior?

- If one tries to analyze an admin area of more specificity than level 2, they get a ValueError.

- [x] Yes
- [x] No

Well, it will technically break for those doing this, but then it wasn't really doing the right thing before. At least noww they'll know it doesn't work.

### DIFF
--- a/api/app/models/common/areas_of_interest.py
+++ b/api/app/models/common/areas_of_interest.py
@@ -1,7 +1,7 @@
 from typing import Annotated, Any, Dict, List, Literal, Optional
 from uuid import UUID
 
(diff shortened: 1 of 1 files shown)


jterry64 (approved):
Thanks!

PR #210, opened 2026-03-04
Add validation for carbon flux

manukala6:
## Pull request checklist

## Pull request type

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

## What is the new behavior?

- Adds post-flight validation for carbon. TCL was refactored while I was doing this PR but I figured we'd do a separate PR to refactor carbon
- 
-

- [ ] Yes
- [x] No


pipelines/carbon_flux/prefect_flows/carbon.py
```diff
@@ -0,0 +1,67 @@
...
+    3: 20,
+    4: 25,
+    5: 30,
+    6: 50,
+    7: 75,
+}
+
+
+def gadm_carbon_flux(tasks, bbox: Optional[Polygon] = None):
+    return compute_carbon_flux(tasks, bbox)
```
danscales: gadm_carbon_flux() and compute_carbon_flux() are identical, not sure why you need both.  Just remove lines 24-27, so you only have gadm_carbon_flux().

pipelines/carbon_flux/stages.py
```diff
@@ -130,35 +30,167 @@ def setup_compute(
...
+        values = sparse_data.data
+
+        # create coordinate dictionary
+        coord_dict = {
+            dim: alerts_count.coords[dim].values[indices[i]]
+            for i, dim in enumerate(dim_names)
+        }
+        coord_dict["value"] = values
+
+        df = pd.DataFrame(coord_dict)
```
danscales: Please  replace lines 154-167 with a call common_stages.py:create_result_dataframe() (as it was before). No need to duplicate code. This will do all these lines, and also the country processing that you had to add at the end of carbon.py:compute_carbon_flux().

pipelines/carbon_flux/prefect_flows/carbon_flow.py
```diff
@@ -1,62 +1,41 @@
...
-
-    mangrove_stock_2000_zarr_uri = "s3://lcl-analytics/zarr/jpl_mangrove_aboveground_biomass_stock_2000/v201902/is_mangrove.zarr/"
-
-    tree_cover_gain_from_height_zarr_uri = "s3://lcl-analytics/zarr/umd_tree_cover_gain_from_height/v20240126/period.zarr/"
-    tree_cover_density_2000_zarr_uri = "s3://lcl-analytics/zarr/umd_tree_cover_density_2000/v1.8/threshold.zarr/"
-
-    result_uri = "s3://lcl-analytics/zonal-statistics/admin-carbon.parquet"
-    funcname = "sum"
+@flow(name="Carbon Flux")
+def gadm_carbon_flux_flow(
```
danscales: Since you changed the name of this function, you need to update the call at line 64 in run_updates.py.

pipelines/carbon_flux/stages.py
manukala6: switched back to common_stages.py:create_result_dataframe()

pipelines/carbon_flux/stages.py
```diff
@@ -132,37 +35,260 @@ def setup_compute(
...
+
+                return pd.Series(
+                    {
+                        "pass": emissions_result,
+                        "sample_gross_emissions": sample_gross_emissions_total,
+                        "validation_gross_emissions": validation_gross_emissions_total,
+                        "detail": "",
+                    }
+                )
+            except Exception as e:
```
danscales: Can you remove this exception catching here? I'm guessing you only put it in to catch divides by zero, but you should catch that above now. It will be much more informative to debug later if you don't catch and hide any unknown exception here.

danscales (commented):
Looking pretty good, but I asked for some important corrections below.  Have you actually run the validator, and does it succeed on the first 20 QC geometries within 1% tolerance? And how long does it take to run?

jterry64 (changes requested):
I think things may have gotten a bit wonky with the refactor and then refactor back. I suggest in my comments you might just start from a fresh branch and copy/paste the functions I mentioned, otherwise you might bring in a lot of unintentional changes. Happy to pair on it with you, I think we could knock it out pretty fast.

PR #229, opened 2026-04-01
Carbon validation

manukala6:
- Adds a QC step to the carbon flux pipeline that validates pipeline output against GEE reference data, mirroring the existing TCL validation.
- Adds qc_against_validation_source and get_carbon_validation_statistics to carbon_flux/stages.py, comparing pipeline emissions/removals totals against GEE at a 2% threshold
- Registers carbon GEE assets in GoogleEarthEngineDatasetRepository
- Extracts _symmetric_relative_difference from TCL stages into common_stages so both pipelines share it
- Adds integration tests covering the full flow with mocked zarrs and the QC validation logic end-to-end with a fake GEE repo


jterry64 (approved):
Looks great now, and got your tests to pass for me! Thanks for all the work on this.

danscales (approved):
Great, glad you were able to switch to the same pattern as the TCL validation, and all the tests are passing!

PR #237, opened 2026-04-09
Fix logging AnalyticsIn.thumbprint instead of service.resource_thumbprint

dmannarino:
Bug just noticed, restricted to a logging statement but could be confusing.


gtempus (approved):
Looks good, @dmannarino! Thank you!

PR #249, opened 2026-04-15
Fix carbon pipeline to put carbon parquet in correct version folder.

danscales:
Fix carbon pipeline to put carbon parquet in correct version folder.

The carbon zarrs are already fixed at using versions v20260327 from the data-lake and generating the same version v20260327 zarrs. So, the carbon parquet and the carbon in the TCL parquet will only be correct for version 1.13 (TCL 2025), which is all we want/need at point.


solomon-negusse (commented):
Looks good, just needs one small change pointing staging otf to v1.13 [here](https://github.com/wri/project-zeno-data-infra/blob/main/api/app/domain/analyzers/carbon_flux_analyzer.py#L26)

jterry64 (approved):
I think that pivot looks good

PR #253, opened 2026-04-24
Adjust TCL pipeline post-processing to allow >= query for carbon

danscales:
Adjust TCL pipeline post-processing to allow >= query for carbon

After calculating the 30/50/75% values for carbon emissions in TCL pipeline, then subtract 50% value from 30% value, and 75% value from 50% value. This allows the query for carbon to be canopy_cover >= 30/50/75%, rather than canopy_cover == 30/50/75%.

The parquet has already been adjusted in this way and saved as:

s3://lcl-analytics/zonal-statistics/tcl/v1.13/admin-tree-cover-loss_v20260424.parquet

### DIFF
--- a/pipelines/tree_cover_loss/stages.py
+++ b/pipelines/tree_cover_loss/stages.py
@@ -314,6 +314,41 @@ def create_result_dataframe(result: xr.DataArray) -> pd.DataFrame:
     return df_pivoted
 
 
+# Temporarily pivot the canopy_cover column, so we can subtract, for each unique set
+# of ids and contextual columns, carbon_emissions_MgCO2e value for 50% canopy from
+# the 30% value, and the 75% value from the 50% value. This means our TCL carbon query
+# can then test for canopy_cover >= 30/50/75, rather than canopy_cover == 30/50/75.
+def unaggregate_carbon_by_canopy_cover(df: pd.DataFrame):
+    id_cols = ['aoi_id', 'aoi_type']
+    context_cols = [
+        'tree_cover_loss_year', 'is_intact_forest', 'tree_cover_loss_driver',
+        'is_primary_forest', 'natural_forest_class'
+    ]
+    group_cols = id_cols + context_cols
+    target_col = 'carbon_emissions_MgCO2e'
+
+    # Pivot the data to make subtraction easy. This aligns the canopy values side-by-side
+    # for every unique group
+    pivoted = df.pivot_table(index=group_cols, columns='canopy_cover', values=target_col).fillna(0)
+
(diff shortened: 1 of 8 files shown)


solomon-negusse (commented):
The pipeline changes LGTM (besides the clarifying the question), @danscales.  The carbon and tree cover loss OTF analyzers don't have to be updated?

pipelines/tree_cover_loss/stages.py
solomon-negusse, line 416: Changing this to `result_df['canopy_cover'] == T` wouldn't have had the same effect as the new function?

danscales: The 30/50/75 computation is this weird computation that not only tests canopy cover threshold, but also adds in tree cover gain from height and mangroves at each level (30/50/75), and gives the exact sum for that threshold value.  But for the pre-calc TCL query, we want to be able to test canopy_cover >= (not ==) for 30/50/75. So, the new function is doing differences between the results for 30/50/75, so queries with >= 30/50/75 will get the correct final value.

danscales:
> The pipeline changes LGTM (besides the clarifying the question), @danscales. The carbon and tree cover loss OTF analyzers don't have to be updated?

I assume the OTF analyzers are already doing the right calculation, since there is bug filed on them. Pipeline changes, of course, don't affect the OTF analyzers, so need to just them because of this change.

pipelines/tree_cover_loss/stages.py
```diff
@@ -314,6 +314,41 @@ def create_result_dataframe(result: xr.DataArray) -> pd.DataFrame:
     return df_pivoted
 
 
+# Temporarily pivot the canopy_cover column, so we can subtract, for each unique set
+# of ids and contextual columns, carbon_emissions_MgCO2e value for 50% canopy from
+# the 30% value, and the 75% value from the 50% value. This means our TCL carbon query
+# can then test for canopy_cover >= 30/50/75, rather than canopy_cover == 30/50/75.
+def unaggregate_carbon_by_canopy_cover(df: pd.DataFrame):
```
jterry64, line 321: It's hard to imagine how this works but I think it makes sense... maybe it's worth a unit test with a fake pandas dataframe just to confirm the logic works how you expect?

jterry64 (changes requested):

pipelines/tree_cover_loss/stages.py
solomon-negusse: Okay, thanks. Amazing that you got this working!

solomon-negusse (approved):

jterry64 (approved):

PR #255, opened 2026-05-11
fix webhook

solomon-negusse:
This fixes a regression issue introduced when we added the ability to trigger analysis pipeline updates from the Prefect Cloud UI. The issue is blocking data-pump webhook triggers and we're having to manually trigger it from the UI weekly.

### DIFF
--- a/pipelines/disturbance/prefect_flows/dist_flow.py
+++ b/pipelines/disturbance/prefect_flows/dist_flow.py
@@ -52,6 +52,9 @@ def dist_alerts_flow(dist_version=None, overwrite=False, is_latest=False) -> lis
     logger = get_run_logger()
     result_uris = []
 
+    is_latest = str(is_latest).lower() == "true"
+    overwrite = str(overwrite).lower() == "true"
+
     if dist_version is None:
         is_latest = True
         dist_version = get_new_dist_version()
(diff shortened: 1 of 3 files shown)


pipelines/disturbance/prefect_flows/dist_flow.py
```diff
@@ -52,6 +52,9 @@ def dist_alerts_flow(dist_version=None, overwrite=False, is_latest=False) -> lis
     logger = get_run_logger()
     result_uris = []
 
+    is_latest = str(is_latest).lower() == "true"
+    overwrite = str(overwrite).lower() == "true"
+
     if dist_version is None:
```
danscales, line 55: I don't understand why you need to do this.  If is_latest and overwrite are truly booleans, then these two lines are noops, right? So, I'm assuming you are handling a case where is_latest and overwrite might be strings? If so, can you put a comment in and describe why that might be happening? (I'm guessing this is happening when called by the webhook.)

solomon-negusse: Good point @danscales, moved these to the entrypoint script and added clarifying comment.
https://github.com/wri/project-zeno-data-infra/pull/255/changes/00ef17136096f8c898652347189e3b4dcfa4c5c8

danscales (approved):
Thanks for adding the comments, etc.

PR #259, opened 2026-05-22
Adding a flow to create a monthly int-dist zarr

danscales:
We create the zarr for the full global from the int-dist tiles, since the COGs are split (tropics/non-tropics).

### DIFF
--- a/pipelines/intdist/create_zarr.py
+++ b/pipelines/intdist/create_zarr.py
@@ -1,0 +1,59 @@
+import numpy as np
+import xarray as xr
+import pandas as pd
+import io
+import boto3
+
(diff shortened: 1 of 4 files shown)


solomon-negusse (approved):
Looks good,  Dan but will need to add the flow name to the list [here](https://github.com/wri/project-zeno-data-infra/blob/main/terraform/prefect.tf#L163) to be able to trigger it from prefect cloud

danscales:
> Looks good, Dan but will need to add the flow name to the list [here](https://github.com/wri/project-zeno-data-infra/blob/main/terraform/prefect.tf#L163) to be able to trigger it from prefect cloud

Great, thanks for letting me know about that needed addition.  Made that change.

PR #260, opened 2026-05-29
PZB-835: Pure refactor of TCL Analyzer code to ease adding TCLF

manukala6:
This PR only refactors the TCL Analyzer code to create some seams that will facilitate testing when we add the fires intersection (in next PR).

## Pull request type

 

 

- [ ] Yes
- [x] No

### DIFF
--- a/api/app/domain/analyzers/tree_cover_loss_analyzer.py
+++ b/api/app/domain/analyzers/tree_cover_loss_analyzer.py
@@ -82,6 +82,11 @@ async def build_query(analytics_in: TreeCoverLossAnalyticsIn) -> DatasetQuery:
     else:
         query.group_bys.append(Dataset.tree_cover_loss)
 
+    # if by fire, return both TCLF and non-fire TCL results by year
+    # TODO: should this be by driver OR by fire?
+    if "fire" in analytics_in.intersections:
+        query.aggregate.datasets.append(Dataset.tree_cover_loss_from_fires)
+
     if analytics_in.canopy_cover is not None:
         query.filters.append(
             DatasetFilter(
--- a/api/app/domain/analyzers/tree_cover_loss_analyzer.py
+++ b/api/app/domain/analyzers/tree_cover_loss_analyzer.py
@@ -120,16 +120,17 @@ class TreeCoverLossAnalyzer(Analyzer):
 
     @nr_agent.function_trace(name="TreeCoverLossAnalyzer.analyze")
     async def analyze(
-        self, analysis: Analysis, query_service: DuckDbPrecalcQueryService | None = None
+        self, analysis: Analysis
     ) -> None:
         if self.input_uris is None:
             raise Exception("Input URIs must be provided for actual analysis")
 
         analytics_in = TreeCoverLossAnalyticsIn(**analysis.metadata)
 
+        
         if analytics_in.aoi.type == "admin":
             results = await self.analyze_admin_areas(
-                analytics_in, query_service=query_service
+                analytics_in
             )
         else:
             results = await self.analyze_otf(analytics_in)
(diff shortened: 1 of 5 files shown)


api/app/domain/analyzers/tree_cover_loss_analyzer.py
```diff
@@ -49,77 +55,114 @@
...
-            )
-
-        if analytics_in.forest_filter == "primary_forest":
-            query.filters.append(
-                DatasetFilter(
-                    dataset=Dataset.primary_forest,
-                    op="=",
-                    value=1,
-                )
+
```
danscales: Can you delete this line.  It just has space on it, which isn't useful and shows up in flake8.

api/app/domain/analyzers/tree_cover_loss_analyzer.py
```diff
@@ -49,77 +55,114 @@
...
             results[Dataset.carbon_emissions.get_field_name()] = np.nan
 
         analysis.result = results
+
+    async def analyze_admin_areas(
+        self,
+        analytics_in: TreeCoverLossAnalyticsIn,
+    ) -> Dict:
+        if self.input_uris is None:
+            raise Exception("Input URIs must be provided for actual analysis")
```
danscales: This check here was already done above in analyze(), line 125, so this is can be removed, right?

danscales (commented):
Seems reasonable, but please take a look at my comments below.

Also, can you update your description if this is not a pure refactor? I may not be understanding, but it seems like you are adding in the analyze_admin_area() for admin AOIs,  which is going to use DuckDB, whereas the original code always did OTF. Is that correct. If so, can you update your description and maybe put in a comment about that?

dmannarino:
> Seems reasonable, but please take a look at my comments below.
> 
> Also, can you update your description if this is not a pure refactor? I may not be understanding, but it seems like you are adding in the analyze_admin_area() for admin AOIs, which is going to use DuckDB, whereas the original code always did OTF. Is that correct. If so, can you update your description and maybe put in a comment about that?

It was certainly not the intent to make such a change, but I may have misunderstood the code. Looking into it, thanks.

dmannarino:
> > Seems reasonable, but please take a look at my comments below.
> > Also, can you update your description if this is not a pure refactor? I may not be understanding, but it seems like you are adding in the analyze_admin_area() for admin AOIs, which is going to use DuckDB, whereas the original code always did OTF. Is that correct. If so, can you update your description and maybe put in a comment about that?
> 
> It was certainly not the intent to make such a change, but I may have misunderstood the code. Looking into it, thanks.

@danscales I think you're incorrect. I believe the results are generated the same way they always were (in the end), with the precalculated tables for admin AOIs. With the old code you have to trace through the ComputeEngine and the handlers to see that's what's going on, with the new code it's just simpler to see that's what's happening. It's true that in the old code the handler path allowed a fallback of using OTF, if that's what you're thinking, but in practice that path would never be taken.

solomon-negusse (approved):
Nice atomic refactor!

gtempus:
This is a great start. What's the next move?
I expected to see the older more general `compute_engine` [replaced](https://github.com/wri/project-zeno-data-infra/blob/9f29eb8ed281b697fe834686c95df1ff69df7213/api/app/routers/land_change/tree_cover_loss/tree_cover_loss.py#L59). Is that coming up in a followup PR?

dmannarino:
> This is a great start. What's the next move? I expected to see the older more general `compute_engine` [replaced](https://github.com/wri/project-zeno-data-infra/blob/9f29eb8ed281b697fe834686c95df1ff69df7213/api/app/routers/land_change/tree_cover_loss/tree_cover_loss.py#L59). Is that coming up in a followup PR?

Yep! Getting rid of the compute_engine is coming in a follow-up PR. It's still needed for the OTF codepath, but can be removed after switching to directly using FloxOTFHandler.

PR #261, opened 2026-05-29
PZB-835: Add TCLF intersection and analysis

manukala6:
## Pull request type
- [x] Feature
Issue Number: [PZB-835](https://gfw.atlassian.net/browse/PZB-835)

## What is the new behavior?
- Computes both TCLF area and non-fire TCL area when requesting fire analysis
- Adds unit tests for TCLF admin area and OTF analyses

- [ ] Yes
- [x] No

[PZB-835]: https://gfw.atlassian.net/browse/PZB-835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/app/domain/models/dataset.py
+++ b/api/app/domain/models/dataset.py
@@ -18,6 +18,7 @@ class Dataset(Enum):
     tree_cover_gain = "tree_cover_gain"
     tree_cover_loss = "tree_cover_loss"
     tree_cover_loss_drivers = "tree_cover_loss_driver"
+    tree_cover_loss_from_fires = "tree_cover_loss_from_fires"
 
     def get_field_name(self):
         DATASET_TO_NAMES = {
@@ -31,6 +32,7 @@ class Dataset(Enum):
             Dataset.tree_cover_gain: "tree_cover_gain_period",
             Dataset.tree_cover_loss: "tree_cover_loss_year",
             Dataset.tree_cover_loss_drivers: "tree_cover_loss_driver",
+            Dataset.tree_cover_loss_from_fires: "tree_cover_loss_from_fires_area_ha",
         }
 
         return DATASET_TO_NAMES[self]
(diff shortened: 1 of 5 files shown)


api/app/domain/models/dataset.py
```diff
@@ -31,6 +32,7 @@ def get_field_name(self):
             Dataset.tree_cover_gain: "tree_cover_gain_period",
             Dataset.tree_cover_loss: "tree_cover_loss_year",
             Dataset.tree_cover_loss_drivers: "tree_cover_loss_driver",
+            Dataset.tree_cover_loss_from_fires: "tree_cover_loss_from_fires_area_ha",
```
jterry64, line 35: OTF might not actually work here, because the Zarr is actually just the year of TCLF, not the TCLF ha (unless you created one). @danscales added some code to the pipeline to read TCLF zarr and then use it to mask the pixel area. You also need to add the zarr to the ZarrDatasetRepository to use in OTF.

@gtempus curious your thoughts, but since this might be tricky I'd say we should just go ahead with the releasing the precalc querying which looks good, and then do OTF as a follow up PR, and just enable this option for now only for admin type.

jterry64 (changes requested):

api/app/domain/models/dataset.py
gtempus: Yes, let's just allow for fires in admin AOIs and follow up for OTF, @jterry64 and @manukala6 (unless TCLF area has already been added).

manukala6: @gtempus @jterry64 There's a TCLF year zarr but not TCLF area zarr. I added a commit that allows only admin area analysis for TCLF and I xfailed the TCLF OTF test

gtempus (approved):
Nice way to stop this early on in the router, @manukala6 💪

jterry64 (approved):

PR #265, opened 2026-06-03
Tclf otf analysis

manukala6:
Enables TCLF OTF analysis. `FloxOTFHandler` is updated to handle the TCLF Zarr, summing TCLF area instead of grouping (like we do with drivers)


dmannarino:
The only thing I am wary of is this comment:
> This assumes that area_hectares is ordered first in Datasets

I'm not sure if which structure that comment is with regard to, but how easy would it be to make this comment obsolete?

danscales:
> The only thing I am wary of is this comment:
> 
> > This assumes that area_hectares is ordered first in Datasets
> 
> I'm not sure if which structure that comment is with regard to, but how easy would it be to make this comment obsolete?

Yes, I also would like to get rid of the reason for this comment.  Is it really required, and if so, can you explain the reason for the required ordering? And if so, given the explanation, hopefully we can figure out a way to express things so that ordering is not required.

manukala6:
@danscales @dmannarino TCLF is aggregated by `area_ha`, so I needed `area_ha` to exist before the `xr.where()`. In [91f6de3](https://github.com/wri/project-zeno-data-infra/pull/265/commits/91f6de30c177780dbd3fb58e2e95b78e0c37016e) I added an if statement to ensure `area_ha` is loaded prior to TCLF, let me know if that is good.

dmannarino:
> > The only thing I am wary of is this comment:
> > > This assumes that area_hectares is ordered first in Datasets
> > 
> > 
> > I'm not sure if which structure that comment is with regard to, but how easy would it be to make this comment obsolete?
> 
> Yes, I also would like to get rid of the reason for this comment. Is it really required, and if so, can you explain the reason for the required ordering? And if so, given the explanation, hopefully we can figure out a way to express things so that ordering is not required.

Claude helped clarify exactly what is going on here:

"Despite the capital "D" suggesting the Dataset enum itself, the comment is actually about the list being iterated — query.aggregate.datasets, the list of Dataset values on the incoming DatasetQuery.aggregate. The loop builds up an xr.Dataset named by one variable at a time, and the TCLF branch reads by["area_ha"]. That key only exists if a previous iteration already inserted it — which happens when Dataset.area_hectares (whose field name is "area_ha", per dataset.py line 25) was processed earlier in the same loop.
So the ordering constraint is on the query's aggregate.datasets list at runtime, not on any dictionary.
The zarr_dataset_repository dictionary
The dict you asked about (the DATASET_TO_URI mapping in ZarrDatasetRepository) is just a key-based URI lookup — it's accessed via dataset_repository.load(ds, ...), which indexes into it by enum key. Its insertion order has no effect on this code path. You can reorder entries in that dict freely; area_hectares does not need to appear first there.
Will it cause issues?
Yes, but only if a query is constructed with tree_cover_loss_from_fires appearing in aggregate.datasets before area_hectares (or without area_hectares at all). In that case by["area_ha"] won't exist yet and you'll get a KeyError from xarray on line 102. The comment is flagging a latent bug rather than describing a robust design — it would be safer to either:

1. Sort query.aggregate.datasets so area_hectares is processed first, or
2. Load area_ha explicitly inside the TCLF branch rather than relying on loop-order side effects, or
3. Assert/validate at the top of _handle that area_hectares precedes tree_cover_loss_from_fires when the latter is present."

IMO, 2 sounds like the best option.

jterry64 (approved):
Sweet, this should work now! I'm not quite understanding @dmannarino's caught issue above though, so maybe I'm not seeing something

dmannarino:
> Sweet, this should work now! I'm not quite understanding @dmannarino's caught issue above though, so maybe I'm not seeing something

@jterry64 Oops, I meant to approve this after latest commits. Maanas addressed it.

dmannarino (approved):

PR #267, opened 2026-06-08
Extend the date range for dist alerts to 2030/1/1

danscales:
Extend the date range for dist alerts to 2027/6/23

We were about to run out, since end point was 2026/6/23 (2000 days from 2020/12/31).


solomon-negusse (approved):
Thanks for catching that!

PR #272, opened 2026-06-16
PZB-965 Adding GADM int-dist alerts area to the int-dist flow.

danscales:
PZB-965 Adding GADM int-dist alerts area to the int-dist flow.

Had to add in 10m-resolution pixel area and GADM zarrs.

The base date for int-dist alerts is 2014/12/31, whereas the base date
for dist-alerts is 2020/12/31.

I didn't try to share more common code with the dist alerts pipeline,
since dist alerts is likely being replaced by int-dist alerts.

We are starting out with no other contextual layers, like grasslands,
natural lands, etc. AJ agreed to start out with the minimum right now.

Converted from country/region/subregion to aoi_id and do GADM rollup
(different from Dist).

I have not added any tests yet, but I have run the full int-dist
pipeline and inspected the parquet.


jterry64 (changes requested):
@danscales it looks good other than some general naming changes, both in code and for column:
- Can you replace usage of "intdist" with "integrated_alerts" in the code, folder names, etc.?
-Can you remove the "intdist" prefix for parquet columns?

jterry64:
@danscales results are also in m^2, not hectares, I think you need to divide by 10000

solomon-negusse (approved):
This looks good to me once Justin's comments are addressed.

danscales:
> @danscales it looks good other than some general naming changes, both in code and for column:
> 
> * Can you replace usage of "intdist" with "integrated_alerts" in the code, folder names, etc.?
>   -Can you remove the "intdist" prefix for parquet columns?

OK, I renamed intdist and intdist_alerts to integrated_alerts everywhere, including in file names.  There were already a few integrated alerts files that were mostly skeleton and not being used, so I removed them before renaming the integrated dist alerts files into their place.

danscales:
> @danscales results are also in m^2, not hectares, I think you need to divide by 10000

Oops, I didn't notice this until after I had run the pipeline again, so I'm just going to check things in as is (without adding the divide by 10000).  As I am saying in my Slack message, can you just update the parquet to divide the area by 10000 and add the divide to the code (or have someone else add the divide by 10000, maybe Maanas when he adds the tests)?  I don't want to run the pipeline again locally, because I am running out of free monthly coiled credits.

PR #276, opened 2026-07-06
Run the integrated_alerts_flow when an intdist-updated-event webhook is received

danscales:
Run the integrated_alerts_flow when an intdist-updated-event webhook is received

This terraform is completely in analogy with the code to handle the dist-updated-event webhook.

I have a parallel change to the datapump that sends the intdist-updated-event webhook when a new intdist version is created on Sunday or the next following day that creates a version.

### DIFF
--- a/terraform/prefect.tf
+++ b/terraform/prefect.tf
@@ -213,6 +213,52 @@ resource "prefect_automation" "run_pipelines_on_dist_update" {
 }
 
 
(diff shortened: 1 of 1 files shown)


jterry64 (approved):
Looks good, same comment as other: wouldn't it be simpler if we just polled every Monday, or is there some logic to deal with?

PR #279, opened 2026-07-16
PZB-1206: Land GHG Inventory pipeline - vegetation

jterry64:
# Add Land GHG Inventory (vegetation) zonal-statistics pipeline

## Summary

Adds a new precompute pipeline that turns the WRI AFOLU GHG-flux model's **vegetation** outputs into GADM admin-level zonal statistics. For every `aoi_id` it produces gross emissions / gross removals / net flux / area, broken down by **land-state class × year (2016–2024)**, written to a versioned parquet.

This is the first of the "land GHG inventory" family; **soil (mineral/organic) is deliberately a separate future pipeline** with its own output parquet.

## What it does

Reads global 30 m zarrs (per-hectare fluxes from the AFOLU flux model + GADM admin + pixel area), converts per-hectare → per-pixel, collapses the model's 95 `land_state_node` codes into 4 MVP vegetation classes, runs a `flox` sparse zonal reduce by `admin × land_state_class × year`, and rolls up to adm0/adm1/adm2 `aoi_id`s.

```mermaid
flowchart TD
  subgraph In["Global 30 m zarrs"]
    V["vegetation zarr<br/>per-ha fluxes + land_state_node<br/>years 2016-2024"]
    A["pixel_area (hectares)"]
    G["GADM adm0 / adm1 / adm2"]
  end
  V --> L["load_data<br/>clip bbox - align to veg grid (1:1, native 30 m)"]
  A --> L
  G --> L
  L --> S["setup_vegetation_compute<br/>collapse land_state 95 to 4<br/>per-ha flux x area to per-pixel cube (+ area_ha)"]
  S --> R["common_stages.compute<br/>flox xarray_reduce(sum), SPARSE_COO<br/>group by admin x land_state_class x year"]
  R --> DF["result dataframe<br/>pivot layers to wide - map class to name<br/>drop 'excluded' - year index to calendar"]
  DF --> RU["rollup_by_gadm_and_convert_to_aoi<br/>adm0 / adm1 / adm2 rows"]
  RU --> P[("admin parquet")]
```

### The land_state collapse

The model encodes each pixel's land-use transition as one of 95 `land_state_node` codes; we bake the code→class map (`CATEGORY_LAND_STATES`, from lookup sheet `v105_20260601`) and relabel before the reduce, which keeps the group space tiny.

```mermaid
flowchart LR
  N["land_state_node<br/>(95 codes)"] --> C{"collapse<br/>LAND_STATE_TO_CATEGORY"}
  C --> TL["tree_loss (44)"]
  C --> TG["tree_gain (6)"]
  C --> TR["trees_remaining_trees (29)"]
  C --> NT["non_trees_remaining_non_trees (15)"]
  C --> EX["excluded (1) -> dropped"]
```

Mangroves count as trees (loss/gain/remaining map through `detailed_class`). The three "mangrove mask" bookkeeping codes share `(mangrove_other, tree)`, so they're assigned **per-code** from their researcher-confirmed meanings: `10000000`→tree_gain, `10100000`→non_trees, `10200000`→tree_loss; only `70000000` ("Not in decision tree") stays excluded.

## Running it

Wired into `run_updates.py` as a new `land_ghg_inventory_update` flow. Scope (`--bbox`) and backend (`--local`) are independent; Coiled is the default.

```mermaid
flowchart TD
  CLI["run_updates --flow land_ghg_inventory_update<br/>--version V [--bbox ...] [--local] [--performance-report x.html]"] --> Q{"--local?"}
  Q -- "no (default)" --> Coiled["Coiled cluster"]
  Q -- yes --> Loc["local dask scheduler"]
  Coiled --> F["land_ghg_inventory_area(version, bbox)"]
  Loc --> F
  F --> B{"bbox set?"}
  B -- "no (global)" --> S3[("s3://.../zonal-statistics/<br/>land_ghg_inventory-vegetation/{version}/...parquet")]
  B -- yes --> LP[("local admin-...-{version}.parquet<br/>(partial run)")]
```

```bash
# global (Coiled)
python -m pipelines.run_updates --flow land_ghg_inventory_update --version global

# one area, on the laptop
python -m pipelines.run_updates --flow land_ghg_inventory_update --version test \
  --bbox "-74,-11.2,-66.5,-7" --local
```

`--performance-report x.html` emits a Dask perf report (Coiled runs only).

## Output schema

| column | notes |
|---|---|
| `aoi_id`, `aoi_type` | GADM adm0/adm1/adm2 |
| `land_state_class` | tree_loss / tree_gain / trees_remaining_trees / non_trees_remaining_non_trees |
| `year` | 2016-2024 |
| `gross_emissions_MgCO2e`, `gross_removals_MgCO2`, `net_flux_MgCO2e`, `area_ha` | `net = emissions + removals`; removals <= 0 |

## Validation

Reproduces the researchers' published parquet:
- **Global: 233 / 235 countries match** within a dual threshold (rel > 2% **and** abs > 1000 MgCO2e). The only two gaps are reference-only structural buckets — `NA` (pixels outside GADM borders, which we drop) and `ATA` (Antarctica, negligible flux).
- **South America, at the individual `land_state` code level: 13 / 14 countries match.** Only Ecuador flags, because Galápagos (~91°W) sits outside the SA bounding box — not a pipeline error.

## Testing

- **Unit** (`test/unit/land_ghg_inventory/`): land_state collapse (counts + representative + mangrove-mask codes), per-pixel cube math & groupbys, result dataframe rollup/mapping.
- **Integration** (`test/integration/land_ghg_inventory/`): runs the real stages clipped to São Tomé & Príncipe and asserts known reference totals + flux-sign sanity.

## QC

Automated in-flow QC was intentionally moved to a **notebook** (`notebooks/land_ghg_inventory_vegetation_qc.ipynb`): point it at a reference parquet + a pipeline output and it compares all countries by `land_state_class` (and can go finer, by `land_state_detailed_class`). The flow itself is just compute → save.

## Notes / out of scope

- **Consolidated 4-category form.** A raw all-95-`land_state` variant was explored but reverted: on South America it was ~2× slower and **transfer-bound** (~40 h aggregate transfer vs ~7 h for the 4-category version, plus spilling), because the 95-way sparse partials blow up the cross-worker combine. Consolidating keeps the global run compute-bound.
- **Soil**, **agriculture**, the `latest` marker, and API serving are follow-ups.
- The old `notebooks/afolu_..._prototype.ipynb` still references the pre-rename module and is not updated here.


jterry64:
@gtempus auto-review:

# PR Review — Land GHG Inventory (vegetation) zonal-statistics pipeline

> Scope note: this PR adds an **offline batch pipeline** (Prefect flow + dask/flox reduce over zarr → parquet). It exposes **no HTTP API, no auth surface, no DB migrations, no async request path**. Review dimensions **2 (API Design)** and most of **3 (OWASP API Security / LRO)** are therefore **Not Applicable**; they are scored N/A rather than padded with speculative findings. Grading reflects dimensions 1, 4, 5, 6 and the applicable slivers of 3.

## Overall Grade: A-

Clean hexagonal separation (pure transform `stages.py`, thin Prefect `@task` adapters, dependencies declared at the flow/CLI edge), domain rules covered by **in-memory fakes rather than mocks**, and validated against an external reference (233/235 countries). The grade is held below A by three real issues: (1) `expected_groups` hardcodes GADM cardinality as magic numbers with **silent truncation** semantics, (2) the orchestration layer (`_parse_bbox`, flow dispatch, skip-if-exists, local/Coiled branching) is **untested**, and (3) the output carries an intentional-but-sharp **NaN-for-zero** data contract that will surprise downstream consumers. None are blockers at zero clients.

## What's Working Well 🟢

1. **Textbook hexagonal split.** `pipelines/land_ghg_inventory/stages.py` is pure domain logic (`collapse_land_state`, `setup_compute`, `create_result_dataframe`) with the only I/O isolated in `load_data` / `_align_to` (the zarr driven-adapter). The Prefect `@task` wrappers in `land_ghg_inventory_tasks.py` are 3-line pass-throughs — orchestration is kept out of the domain:
   ```python
   @task
   def vegetation_result_dataframe(reduced: xr.DataArray) -> pd.DataFrame:
       return stages.vegetation_result_dataframe(reduced)
   ```
2. **Dependencies at the edge.** Zarr URIs live in `globals.py` and are imported only in the flow module (`land_ghg_inventory_flow.py`), not reached into from the domain — matching CLAUDE.md's "declare dependencies as far out on the edge as possible."
3. **Fakes, not mocks, in unit tests.** `test/unit/land_ghg_inventory/conftest.py::synthetic_datasets` builds a real in-memory 2×2 xarray scene and runs it through the *real* reduce; the unit tests assert domain rules (per-pixel `flux × area`, `net = emis + removals`, 95→category collapse counts, rollup) with **zero network coupling**. Live-data coupling is correctly quarantined in `test/integration/`.
4. **Cluster lifecycle is cleaned up.** `run_updates.py` shuts the Coiled client down in a `finally`, so a failed reduce doesn't leak the cluster:
   ```python
   finally:
       if dask_client:
           dask_client.shutdown()
   ```
5. **Bounded-context hygiene.** The pipeline is a self-contained module reusing the shared kernel (`common_stages`/`common_tasks`) rather than forking it; soil is explicitly deferred to a separate pipeline instead of being crammed in.
6. **Scope/backend decoupling in the CLI.** `--bbox` (what to reduce) and `--local` (where it runs) are independent, and `--performance-report` is correctly gated to Coiled-only runs.

## Issues to Address 🟡

1. **`expected_groups` hardcodes GADM cardinality as magic numbers with silent-truncation semantics.** (`land_ghg_inventory_flow.py:24-30`)
   ```python
   expected_groups = (
       np.arange(999),  # country iso codes
       np.arange(86),   # region codes
       np.arange(854),  # subregion codes
       np.array([0, 1, 2, 3, 4]),
       np.arange(9),
   )
   ```
   **Why it matters:** flox drops any group label outside `expected_groups`. If GADM ever emits a subregion index ≥ 854 or a country code ≥ 999 (adm2 counts routinely exceed 854 for large countries depending on the coding scheme), those pixels are **silently discarded** — no error, just quietly missing/low totals. `854` in particular looks like a copy-pasted "big enough" number (it happens to be Burkina Faso's ISO code). The reduce reproduces the reference *today*, so the ranges are currently sufficient, but the coupling is invisible and unguarded.
   **Fix:** derive the bounds from the data (or the GADM spec) and assert nothing is dropped:
   ```python
   COUNTRY_MAX, REGION_MAX, SUBREGION_MAX = 1000, 90, 900  # from GADM v4.1 spec
   # after the reduce, guard against silent truncation:
   observed = int(country_layer.max().compute())
   if observed >= COUNTRY_MAX:
       raise ValueError(f"country code {observed} exceeds expected_groups bound")
   ```

2. **Intentional NaN-for-zero in the output is an undocumented data contract.** `tree_gain` gross-emissions rows come through as `NaN` (not `0.0`) at leaf adm2 rows, because the `SPARSE_COO` reduce stores no explicit zeros. It's intentional (you preferred it), but it lives only in a code comment.
   **Why it matters:** any downstream consumer that sums, joins, or serializes these columns will propagate/di­verge on NaN (`sum` skips it, but arithmetic, equality, JSON, and many DB loaders don't). For a parquet that *is* the product's contract, "sometimes 0, sometimes NaN, same meaning" is a footgun.
   **Fix:** document it in the schema (README/data dict) **and** make it explicit at the boundary, e.g.
   ```python
   # in create_result_dataframe, before return:
   df[MEASURES] = df[MEASURES].where(df[MEASURES].notna(), other=pd.NA)  # keep NaN, but declared
   ```
   or fill zeros for the measure columns if consumers expect dense — the key is to *decide and document*, not leave it emergent.

3. **The orchestration layer has no tests.** `_parse_bbox`, the `version`-required validation, the `local`→skip-cluster branch, the `bbox`→local-path branch, and `performance_report` gating are all untested; only the pure `stages` functions are covered.
   ```python
   def _parse_bbox(bbox):
       minx, miny, maxx, maxy = (float(v) for v in str(bbox).split(","))
       return box(minx, miny, maxx, maxy)   # ValueError on 3 or 5 values; no min<max check
   ```
   **Why it matters:** these are cheap pure-logic branches that decide *where results are written* and *whether a cluster spins up* — exactly the kind of bug that silently writes a partial run to the global S3 path or bills a cluster you meant to skip. They're trivially unit-testable without network.
   **Fix:** add a `test_run_updates.py` covering `_parse_bbox` (valid, `None`, malformed → clear error, min>max → error) and the dispatch (`version` required for the two flows; `bbox` only threaded to the land-GHG flow).

## Nits 🔵

- `create_result_dataframe(reduced, class_names)` still takes a `class_names` param though only `VEGETATION_CATEGORIES` is ever passed — leftover generality from the reverted multi-component design; inline or drop it.
- `run_land_ghg_inventory_update(version, overwrite, is_latest, bbox)` accepts `is_latest` but never uses it (signature parity with sibling flows). Fine, but a `# unused` note or `**_` would be clearer.
- New `run_updates` params (`bbox=None, local=False, performance_report_path=None`) and `_parse_bbox` lack type hints, unlike the rest of the module.
- `common_stages.compute` uses `print("Starting reduce")` rather than the Prefect logger (relies on `log_prints=True`). Pre-existing, but worth a structured-log pass eventually.
- `# flake8: noqa: E501` is file-wide in `land_state_categories.py`; it also silences E501 on the real code (`classify`), not just the baked data. Narrow it if you care.
- `logger.error("Analysis failed.")` drops structured context (flow name, version, bbox); the traceback survives via re-raise, but a structured field set would help ops.

## Summary Table

| Area | Assessment |
|---|---|
| Architecture / DDD (hexagonal, deps at edge, bounded context) | ✅ |
| API Design (REST/LRO/OpenAPI) | N/A — no API in PR |
| Security — OWASP Web/API Top 10 | ✅ (no attack surface; minor cost/validation notes) |
| Performance & Reliability | ⚠️ (no reduce timeout; memory-bound; retries×cost) |
| Test Quality (fakes not mocks, no net in unit) | ✅ domain / ⚠️ orchestration untested |
| Code Quality (types, logging, magic numbers) | ⚠️ (`expected_groups` magic numbers; NaN contract) |

## Blast Radius Estimate 💥

### Methodology

| Score | Probability | Impact |
|---|---|---|
| 1 | Very unlikely (needs multiple unlikely conditions) | Cosmetic / no data effect |
| 2 | Unlikely under current usage | Minor; easily corrected, caught by QC |
| 3 | Plausible within a few runs / GADM updates | Wrong numbers in a partition; recompute needed |
| 4 | Likely without a fix | Silent bad data reaches a consumer |
| 5 | Near-certain | Broad silent corruption / large cost |

**Blast Radius Score = Probability × Impact** · Thresholds: 🟢 ≤ 6 · 🟡 7–12 · 🔴 ≥ 13

### Risk Matrix

```mermaid
%%{init: {"theme": "base", "themeVariables": {"quadrant1Fill": "#ffd6d6", "quadrant2Fill": "#ffe8cc", "quadrant3Fill": "#d9f2d9", "quadrant4Fill": "#fff5cc"}}}%%
quadrantChart
    title Blast Radius - Probability vs Impact
    x-axis Low Probability --> High Probability
    y-axis Low Impact --> High Impact
    quadrant-1 Watch
    quadrant-2 Fix Before First Client
    quadrant-3 Accept / Document
    quadrant-4 Monitor at Launch
    expected_groups silent truncation: [0.35, 0.85]
    NaN-for-zero data contract: [0.85, 0.38]
    Orchestration layer untested: [0.55, 0.35]
    Integration test external coupling: [0.48, 0.3]
    Memory OOM at larger scope: [0.35, 0.52]
    Unbounded run cost - no timeout: [0.3, 0.48]
```

### Risk Register

| # | Risk | Prob | Impact | Score | OWASP / CWE | Verdict |
|---|---|---|---|---|---|---|
| 1 | `expected_groups` silently drops out-of-range GADM codes | 2 | 5 | 🟡 10 | CWE-682 / CWE-20 | Guard before first consumer |
| 2 | NaN-for-zero undocumented output contract | 5 | 2 | 🟡 10 | CWE-1078 (data quality) | Document + decide |
| 3 | Orchestration logic (`_parse_bbox`, dispatch) untested | 3 | 2 | 🟢 6 | CWE-1120 | Add unit tests |
| 4 | Integration test couples to live external S3 + hardcoded totals | 3 | 2 | 🟢 6 | — | Accept / mark network |
| 5 | Memory OOM at larger group space / bbox | 2 | 3 | 🟢 6 | CWE-400 / API4:2023 | Monitor; documented |
| 6 | No reduce timeout; `retries=2` multiplies cost of a hung/failing run | 2 | 3 | 🟢 6 | CWE-400 / API4:2023 | Monitor at launch |

### Overall Blast Radius Rating

| Scenario | Rating | Rationale |
|---|---|---|
| Today (zero clients) | 🟢 | Internal batch job; worst case is a re-runnable bad parquet or wasted Coiled spend, and the QC notebook is a human gate before anything is trusted. No external surface. |
| First client, no fixes | 🟡 | Once a downstream consumer (API or partner) reads the parquet, the NaN contract (#2) and any silent truncation (#1) become *their* wrong numbers, with no error to trace back to. |
| First client, critical fixes applied | 🟢 | With #1 guarded and #2 documented/decided, the remaining risks are cost/ops-only and monitorable. |

### Required Fixes Before First Client

#### 1. Guard `expected_groups` against silent truncation (🟡 10)

Problem:
```python
expected_groups = (np.arange(999), np.arange(86), np.arange(854), ...)
# any admin code >= these bounds is dropped from the reduce with no error
```
Fix:
```python
# assert the observed admin codes fit before trusting the result
for layer, bound, name in [(country, 999, "country"),
                           (region, 86, "region"),
                           (subregion, 854, "subregion")]:
    hi = int(layer.max().compute())
    if hi >= bound:
        raise ValueError(f"{name} code {hi} exceeds expected_groups bound {bound}")
```
**Why it matters at scale:** a GADM version bump or a coding-scheme change can push a code past the bound, and the failure mode is *not* a crash — it's a country or municipality quietly missing tonnes of CO₂e from the totals. At one consumer that's a silently wrong national inventory number with no signal to investigate; the guard converts a silent data-integrity bug into a loud, actionable failure.

#### 2. Make the NaN-for-zero contract explicit (🟡 10)

Problem:
```python
# tree_gain gross_emissions arrives as NaN, not 0.0, only explained in a comment
```
Fix — pick one and write it in the schema doc:
```python
# Option A (keep NaN, declared): document "measure = NaN means structurally zero"
# Option B (dense): df[MEASURES] = df[MEASURES].fillna(0.0)
```
**Why it matters at scale:** the parquet *is* the API for this product. A consumer computing `emissions / area`, joining on a measure, or loading into a store with a non-nullable float column will break or produce garbage the first time it hits a NaN cell — and because most rows are populated, it'll pass casual testing and fail in production. Deciding the contract now (and documenting it) is far cheaper than reconciling two interpretations across teams later.

---

**Net:** strong, well-architected pipeline that reproduces the reference data and is genuinely test-covered where it counts. Ship-able today given zero clients; before the first downstream consumer, guard the `expected_groups` bounds and pin down the NaN contract.

solomon-negusse (approved):
Looks good! Just one small thing - can you add the flow name here so it's surfaced in the prefect cloud UI @jterry64 
https://github.com/wri/project-zeno-data-infra/blob/main/terraform/prefect.tf#L163

PR #281, opened 2026-07-27
PZB-1208: add agriculture ghg inventory pipeline; clean separation of components

solomon-negusse:
## Summary
- Add the Land GHG inventory **agriculture** component (`land_ghg_inventory_agriculture` flow): cropland + livestock emissions, admin-only zonal stats (no `land_state_class` or year axis — a single static snapshot), rolled up to `aoi_id`.
- Refactor the existing vegetation logic out of the old combined `land_ghg_inventory_area` flow into its own `land_ghg_inventory_vegetation` subflow (renamed `stages.py` → `vegetation_stages.py`), so `land_ghg_inventory_area` is now a thin parent that dispatches to per-component subflows (`vegetation`, `agriculture`) via a `component` arg, each writing its own parquet.
- Wire the new component through to Prefect Cloud: `run_updates.py` gains `land_ghg_inventory_vegetation_update` and `land_ghg_inventory_agriculture_update` as selectable `UpdateFlow` entries (in addition to `land_ghg_inventory_update`, which still runs all components).

### Output schemas

Also documented inline in `agriculture_stages.py` / `vegetation_stages.py`.

**Agriculture** — one row per `aoi_id` x `category`:

| column | type | notes |
|---|---|---|
| `aoi_id` | str | admin unit, e.g. `"BRA"`, `"BRA.1"`, `"BRA.1.1"` |
| `aoi_type` | str | always `"admin"` |
| `category` | str | `"cropland"` \| `"livestock"` |
| `gross_emissions_MgCO2e` | float | summed emissions for `aoi_id` x `category` |

### DIFF
--- a/pipelines/land_ghg_inventory/agriculture_stages.py
+++ b/pipelines/land_ghg_inventory/agriculture_stages.py
@@ -4,6 +4,13 @@ Sums already-absolute per-pixel cropland + livestock emissions grouped by admin
 unit x category (cropland / livestock) only (no land_state_class, no year — a
 single static snapshot), then rolls up to aoi_id. The reduce and GADM roll-up are
 reused from ``pipelines.prefect_flows.common_stages``.
+
+Output parquet schema (one row per aoi_id x category)::
+
+    aoi_id                    str    admin unit, e.g. "BRA", "BRA.1", "BRA.1.1"
+    aoi_type                  str    always "admin"
+    category                  str    "cropland" | "livestock"
+    gross_emissions_MgCO2e    float  summed emissions for aoi_id x category
 """
 
 from typing import Optional, Tuple
--- a/pipelines/land_ghg_inventory/agriculture_stages.py
+++ b/pipelines/land_ghg_inventory/agriculture_stages.py
@@ -1,9 +1,9 @@
 """Zonal-statistics stages for Land GHG inventory agriculture emissions.
 
 Sums already-absolute per-pixel cropland + livestock emissions grouped by admin
-unit only (no land_state_class, no year — a single static snapshot), then rolls up
-to aoi_id. The reduce and GADM roll-up are reused from
-``pipelines.prefect_flows.common_stages``.
+unit x category (cropland / livestock) only (no land_state_class, no year — a
+single static snapshot), then rolls up to aoi_id. The reduce and GADM roll-up are
+reused from ``pipelines.prefect_flows.common_stages``.
 """
 
 from typing import Optional, Tuple
@@ -23,10 +23,10 @@ from pipelines.prefect_flows.common_stages import (
     rollup_by_gadm_and_convert_to_aoi,
 )
 
-# canonical measure -> source variable in the agriculture zarr
+# canonical category -> source variable in the agriculture zarr
 AGRICULTURE_SOURCE_VARS = {
-    "cropland_emissions": "cropland_emissions",
-    "livestock_emissions": "livestock_emissions",
+    "cropland": "cropland_emissions",
+    "livestock": "livestock_emissions",
 }
 AGRICULTURE_ZARR_GROUP = "pipeline"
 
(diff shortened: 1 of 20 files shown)


solomon-negusse:
@jterry64 - just remembered this is missing the zarr creation step. BTW, turned out resampling of the ag data was pretty expensive to do on the fly so opted for one time zarr creation. I'll add that tomorrow but rest of the pipeline is good.

pipelines/land_ghg_inventory/agriculture_stages.py
```diff
@@ -0,0 +1,86 @@
...
+from pipelines.prefect_flows.common_stages import (
+    rollup_by_gadm_and_convert_to_aoi,
+)
+
+# canonical category -> source variable in the agriculture zarr
+AGRICULTURE_SOURCE_VARS = {
+    "cropland": "cropland_emissions",
+    "livestock": "livestock_emissions",
+}
+AGRICULTURE_ZARR_GROUP = "pipeline"
```
jterry64, line 38: why is the agricultural zarr group called pipeline? And is it actually a zarr group or dataset with two vars?

pipelines/test/integration/land_ghg_inventory/test_agriculture_flow.py
```diff
@@ -0,0 +1,54 @@
...
+def test_stp_reproduces_reference_totals():
+    datasets = agriculture_stages.load_agriculture(
+        land_ghg_inventory_agriculture_zarr_uri,
+        country_zarr_uri,
+        region_zarr_uri,
+        subregion_zarr_uri,
+        bbox=STP_BBOX,
+    )
+    expected_groups = (
+        np.arange(999),
```
jterry64: I think I added the expected group counts for the GADM layers to the globals, I know we probably already hardcode them all over tests but might be good to use those going forward

jterry64 (approved):
LGTM other than a couple minor comments! And yes if post the Zarr part I'll look tomorrow. If you could.

It'd also be great if you could include the description of the PR (and maybe we should include this is the docs/comments?) the schema of the output parquet. One thing we also might want to add when we do the soil is an option in the CLI to only trigger vegetation/soil/agriculture for a run, so if one output has a QC issue we don't need to always rerun all of them.

pipelines/land_ghg_inventory/agriculture_stages.py
solomon-negusse: That's [zarr group](https://zarr.readthedocs.io/en/latest/user-guide/groups/) that gives a clean way of storing multiple arrays in a single zarr store. Named this pipeline (as opposed to otf) based on where it's used. TCL has both groups now, for example, with 4k OTF chunk size which is 2.5x more performant than 10k for small AOIs. THis is just in prep of OTF and there isn't much cost to adding this if we end up not doing oTF.

pipelines/test/integration/land_ghg_inventory/test_agriculture_flow.py
solomon-negusse: Good catch, updated.

solomon-negusse:
Thanks for the review @jterry64 

Added zarr creation step https://github.com/wri/project-zeno-data-infra/pull/281/changes/273e6034d0baafefa68ff7162b9a2b8855a2ec37

>One thing we also might want to add when we do the soil is an option in the CLI to only trigger vegetation/soil/agriculture for a run, so if one output has a QC issue we don't need to always rerun all of them.

Already added this option - saw it coming :-) 

Parquet schemas added in PR comment and code https://github.com/wri/project-zeno-data-infra/pull/281/changes/429eed85d8084a6f3cfaa26d8bd86669d1ce784b

pipelines/land_ghg_inventory/vegetation_stages.py
```diff
@@ -4,6 +4,20 @@
 grouped by admin unit x land_state_class x year, then rolls up to aoi_id. The reduce
 and GADM roll-up are reused from ``pipelines.prefect_flows.common_stages``. Soil is a
 separate pipeline with its own output parquet.
+
```
jterry64, line 7: Yes, this is great!

jterry64:
> Thanks for the review @jterry64
> 
> Added zarr creation step [273e603](https://github.com/wri/project-zeno-data-infra/commit/273e6034d0baafefa68ff7162b9a2b8855a2ec37)
> 
> > One thing we also might want to add when we do the soil is an option in the CLI to only trigger vegetation/soil/agriculture for a run, so if one output has a QC issue we don't need to always rerun all of them.
> 
> Already added this option - saw it coming :-)
> 
> Parquet schemas added in PR comment and code [429eed8](https://github.com/wri/project-zeno-data-infra/commit/429eed85d8084a6f3cfaa26d8bd86669d1ce784b)

Nice, still look goods to me!

PR #283, opened 2026-07-28
Fix mypy/flake8 issues and add CORS middleware

jterry64:


gtempus (approved):
Thank you for removing that dead code, @jterry64 💪

PR #286, opened 2026-07-29
PZB-1207: add lgms soils pipeline and API

solomon-negusse:
## Summary

Adds two new components to the `land_ghg_inventory` pipeline — `mineral_soil` and `organic_soil` — alongside the existing `vegetation` and `agriculture` components. Each reads its own source zarr, converts per-hectare fluxes to per-pixel totals, aggregates to GADM admin units, and writes its own parquet, following the same pattern already established for `vegetation`/`agriculture`.

## New parquet outputs

**Mineral soil** — `s3://lcl-analytics/zonal-statistics/land_ghg_inventory-mineral_soil/{version}/admin-land_ghg_inventory-mineral_soil.parquet`

One row per `aoi_id`. A single static snapshot (the 2015–2020 SOC change interval) — no year axis:

| column | type | notes |
|---|---|---|
| `aoi_id` | str | e.g. `"BRA"`, `"BRA.1"`, `"BRA.1.1"` |
| `aoi_type` | str | always `"admin"` |
| `gross_emissions_MgCO2e` | float | |
| `gross_removals_MgCO2` | float | |
| `net_flux_MgCO2e` | float | |
| `area_ha` | float | |

**Organic soil** — `s3://lcl-analytics/zonal-statistics/land_ghg_inventory-organic_soil/{version}/admin-land_ghg_inventory-organic_soil.parquet`

One row per `aoi_id` x `interval_end_year`. Two native 5-year blocks are persisted as-is (`2020` covers 2016–2020, `2024` covers 2021–2024) — not broadcast to annual figures:

| column | type | notes |
|---|---|---|
| `aoi_id` | str | |
| `aoi_type` | str | always `"admin"` |
| `interval_end_year` | int | `2020` \| `2024` |
| `gross_emissions_MgCO2e` | float | burned + drained, summed |
| `area_ha` | float | restricted to organic-soil-extent pixels |

## Downstream API payload

These parquets are read directly by the `land_ghg_inventory` analytics endpoint (see the companion API PR), which serves each one as a column-oriented table under `result`, alongside the existing `vegetation`/`agriculture` tables:

```json
{
  "data": {
    "result": {
      "vegetation": {
        "aoi_id": ["BRA.1", "BRA.1"],
        "aoi_type": ["admin", "admin"],
        "land_state_class": ["tree_loss", "tree_gain"],
        "year": [2016, 2016],
        "gross_emissions_MgCO2e": [438480933.0, 0.0],
        "gross_removals_MgCO2": [-29404371.0, -3213337.0],
        "net_flux_MgCO2e": [409076565.0, -3213337.0],
        "area_ha": [1086087.0, 518423.0]
      },
      "agriculture": {
        "aoi_id": ["BRA.1"],
        "aoi_type": ["admin"],
        "category": ["cropland"],
        "gross_emissions_MgCO2e": [123234500000.0]
      },
      "mineral_soil": {
        "aoi_id": ["BRA.1"],
        "aoi_type": ["admin"],
        "gross_emissions_MgCO2e": [4533948.0],
        "gross_removals_MgCO2": [-2740651.0],
        "net_flux_MgCO2e": [1793298.0],
        "area_ha": [8509086.0]
      },
      "organic_soil": {
        "aoi_id": ["BRA.1", "BRA.1"],
        "aoi_type": ["admin", "admin"],
        "interval_end_year": [2020, 2024],
        "gross_emissions_MgCO2e": [11.6, 313515.6],
        "area_ha": [8509086.0, 8509086.0]
      }
    },
    "metadata": { "aoi": { "type": "admin", "ids": ["BRA.1"] } },
    "message": "Analysis completed successfully.",
    "status": "saved"
  },
  "status": "success"
}
```

Each table's columns map 1:1 to its parquet's schema. `mineral_soil` and `organic_soil` are flat, top-level siblings of `vegetation`/`agriculture` — not nested under a parent `soil` key — since each has an independent schema.

## Key decisions

- **Two components, not one.** Mineral and organic soil have different measures (3 vs. 1) and different temporal shapes (no year axis vs. 2 native blocks), so merging them into a single `soil` parquet with a category discriminator would have produced a sparse/ragged table. Kept as two independent components/parquets, matching the existing one-component-per-source-zarr precedent.
- **No broadcasting to annual years at write time.** Both sources are static/blocky at the source (a single SOC interval; two 5-year organic-soil blocks). Mapping those onto vegetation's 9 annual years is a derivative join, not a fact worth baking into storage — persisted at native resolution, left for downstream consumers to broadcast.
- **`area_ha` is masked to organic-soil extent.** The organic soil source zarr carries an explicit `organic_soil` extent variable; `area_ha` is computed as `pixel_area` restricted to that mask, not total admin-unit area.

## Verification

- New unit tests (synthetic fixtures) for both components' setup/compute and result-dataframe logic.
- New integration tests against the real global zarrs, clipped to São Tomé & Príncipe.
- Cross-checked mineral_soil and organic_soil outputs against an independently-produced reference LULUCF table across a broad country spread (islands, large countries, high/low latitude) — emissions/removals/net flux match closely for the large majority of countries checked; a handful of countries show a real, still-unexplained emissions gap in organic_soil, tracked separately as a follow-up rather than blocking this change.

### DIFF
--- a/api/app/domain/analyzers/land_ghg_inventory_analyzer.py
+++ b/api/app/domain/analyzers/land_ghg_inventory_analyzer.py
@@ -16,6 +16,14 @@ VEGETATION_MEASURES = (
     "area_ha",
 )
 
+# mineral soil measures returned per aoi_id (single static snapshot, no year axis)
+MINERAL_SOIL_MEASURES = (
+    "gross_emissions_MgCO2e",
+    "gross_removals_MgCO2",
+    "net_flux_MgCO2e",
+    "area_ha",
+)
+
 INPUT_URIS = {
     Environment.staging: {},
     Environment.production: {
@@ -26,6 +34,14 @@ INPUT_URIS = {
         "admin_agriculture_results_uri": (
             "s3://lcl-analytics/zonal-statistics/land_ghg_inventory-agriculture/"
             "v20260727/admin-land_ghg_inventory-agriculture.parquet"
+        ),
+        "admin_mineral_soil_results_uri": (
+            "s3://lcl-analytics/zonal-statistics/land_ghg_inventory-mineral_soil/"
+            "v20260729/admin-land_ghg_inventory-mineral_soil.parquet"
+        ),
+        "admin_organic_soil_results_uri": (
+            "s3://lcl-analytics/zonal-statistics/land_ghg_inventory-organic_soil/"
+            "v20260729/admin-land_ghg_inventory-organic_soil.parquet"
         ),
     },
 }
@@ -41,7 +57,13 @@ class LandGHGInventoryAnalyzer(Analyzer):
         land_state_class x year.
       - "agriculture": a coarse snapshot of gross emissions by category
         (cropland) only - no year, removals, net flux, or area.
-    "soil" is added as a further key later."""
+      - "mineral_soil": gross emissions / removals / net flux / area by
+        aoi_id only - a single static snapshot (the 2015-2020 SOC change
+        interval), no year axis.
+      - "organic_soil": gross emissions / area by aoi_id x interval_end_year
+        (2020 or 2024 - the zarr's native 5-year block labels, covering the
+        2016-2020 and 2021-2024 vegetation periods respectively), not
+        broadcast to annual years."""
 
     def __init__(
         self,
@@ -58,11 +80,18 @@ class LandGHGInventoryAnalyzer(Analyzer):
 
         analytics_in = LandGHGInventoryAnalyticsIn(**analysis.metadata)
         aoi_ids = analytics_in.aoi.ids
-        vegetation, agriculture = await asyncio.gather(
+        vegetation, agriculture, mineral_soil, organic_soil = await asyncio.gather(
             self.analyze_vegetation(aoi_ids),
             self.analyze_agriculture(aoi_ids),
+            self.analyze_mineral_soil(aoi_ids),
+            self.analyze_organic_soil(aoi_ids),
         )
-        analysis.result = {"vegetation": vegetation, "agriculture": agriculture}
+        analysis.result = {
+            "vegetation": vegetation,
+            "agriculture": agriculture,
+            "mineral_soil": mineral_soil,
+            "organic_soil": organic_soil,
+        }
 
     async def analyze_vegetation(self, aoi_ids) -> Dict[str, Any]:
         columns = ("aoi_id", "land_state_class", "year") + VEGETATION_MEASURES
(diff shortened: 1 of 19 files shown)


jterry64 (changes requested):
Awesome! A couple questions and thoughts @solomon-negusse:

- Are you sure mineral soil is only 2015-2020? In the mockup they included it annualized across 2015-2024, but the mockup may have been wrong.
- Here we return the periods and the emissions/removals. But I thought the unit was CO2e per year, not gross emissions/removals over the time period? Would it make sense to just annualize it here to make that more understandable, or do you think that should be handled by the charts API?
- Do you think it's worth a dask.persist on re-used datasets between flows, like GADM layers?

solomon-negusse:
Thanks for the review @jterry64. 

> Are you sure mineral soil is only 2015-2020? In the mockup they included it annualized across 2015-2024, but the mockup may have been wrong.

Yeah, they dropped the other time periods and results match qc data. Here is a note from David:
_We're only using the third interval of change (which represents change between 2010/2015 and 2015/2020). The values for this period are used for all years of vegetation data. In the zarr, this is idx=3 (fourth time slice in the zarr), per https://github.com/wri/AFOLU_GHG_flux_model/blob/9318e77c0cdcd11def7c876da598c110842545d1/src/synthesis/scripts/1_LULUCF_flux_summation.py#L16_

>Here we return the periods and the emissions/removals. But I thought the unit was CO2e per year, not gross emissions/removals over the time period? Would it make sense to just annualize it here to make that more understandable, or do you think that should be handled by the charts API?

Yes, it's per year constant over the 5-year interval (adding unit would remove that ambiguity). Had few reasons to return it as interval data vs annualized:
- Main one was efficiency. This blows up parquet size by 4.5x (organic soil) and 9x (mineral soil), which also means every downstream query scans that much more redundant data.
- Annualizing it would hide that this was from a model that outputs result in 5-year intervals
- It make assumption about how downstream consumers choose to process the raw data (maybe would want to interpolate differently?)

With the first point being the main reason, what about doing the broadcasting at the API level? Perhaps annualizing it would make it easier for the agent to do accurate analysis and we can keep the interval_end_year to signal that it's originally in 5-year blocks. What do you think @jterry64?

> Do you think it's worth a dask.persist on re-used datasets between flows, like GADM layers?

Not too optimistic about the impact given that we've been iterating on each one individually and we don't run this often (DIST w all the contextual layer runs would be an excellent candidate in comparison). But it's the better way to do it overall.

jterry64 (approved):
@solomon-negusse thanks for the responses, switching to approved! Yeah if we annualize that could definitely be broadcasted after the parquet read so keep the parquet small. I definitely get what you mean by being less opinionated to downstream consumers, but it's also possible the agent (or the downstream consumers) would be confused how to use the data. Maybe start with this and possibly have both options available in the future?

PR #289, opened 2026-08-17
Added few tasks that happen for int-alerts only on the 1st Sunday of month

danscales:
Added few tasks that happen for int-alerts only on the 1st Sunday of month

The steps are:
 - Copying the new int-dist zarr to s3://gnw-monitoring-data bucket for GNW self-hosted monitoring system (with a data version, so it doesn't overwrite existing zarr).
 - Registering intdist_tropics.tif COG created monthly as GEE asset in forma-250 project. This is for use by Waginingen for creating Drivers of Integrated Alerts (COG was already uploaded to gs://data-api-gee-assets/gfw_integrated_dist_alerts by export_to_gee Data API option.)
 - Copying the intdist_tropics.tif COG to gcs://wri-lcl-integrated-alerts bucket, and the register the COG also as a landandcarbon GEE asset.

I made the 1st Sunday steps as a separate task that runs in parallel with the remaining parquet computation, after the int-dist zarr has been created.

I added some new utils:

 - copy_s3_directory(): written by Claude, recursively copies a large set of files (like zarr folder) quite efficiently.
 - register_gee_asset(): mostly written by me, for registering a public GEE asset in forma-250 or landandcarbon with MODE pyramidingPolicy.
 - transfer_s3_to_gcs(): mostly written by Claude, using 'gcloud transfer jobs create'. Not super-fast for large files, because of AWS egress bandwith limiting.

Claude created a bunch of unit tests that seem reasonable. I also tested running the actual int-dist pipeline from my local machine on the first Sunday of August version v20260802 (but modifying the code so the existing zarr and parquet were not changed).


pipelines/integrated_alerts/create_zarr.py
```diff
@@ -67,4 +105,48 @@ def create_zarr(version, overwrite=False) -> str:
         decoded_alert_data.to_zarr(zarr_uri, mode="w")
         print("Done to_zarr")
 
-    return zarr_uri
+    return zarr_uri, _is_first_sunday_week(version)
+
+
+def first_sunday_processing(version, zarr_uri) -> None:
```
solomon-negusse: I'd recommend naming this `register_intdist_tropics_gee_assets` or something that describes what it does - we have that scheduler function for that info.

pipelines/integrated_alerts/create_zarr.py
```diff
@@ -30,7 +68,7 @@ def create_zarr(version, overwrite=False) -> str:
     )
 
     if s3_uri_exists(f"{zarr_uri}/zarr.json") and not overwrite:
-        return zarr_uri
+        return zarr_uri, False
```
solomon-negusse, line 38: I'd suggest surfacing `_is_first_sunday_week` check outside this function and inside the flow after zarr creation. That way the orchestration is visible at the top flow and we get separation of concern here.

pipelines/integrated_alerts/create_zarr.py
solomon-negusse: Also would be great for this function and its helpers to be in a separate module

solomon-negusse (commented):
I left some comments @danscales  - not blockers but I think would help with making this addition fairly self-contained and easier to maintain.

pipelines/integrated_alerts/create_zarr.py
danscales: OK, I moved all the relevant code to a separate file/module extra_processing.py.  I renamed the function to extra_processing_tasks(). I didn't rename it to register_intdist_tropics_gee_assets() or something like that, because there are at least two distinct tasks going on here (copying new zarr to a GNW monitoring bucket, and registering several GEE assets, including a needed copy of the COG file).

pipelines/integrated_alerts/create_zarr.py
danscales: Yes, good suggestion! That combined with the other suggestions keeps the create_zarr() almost completely clean and unchanged. It now just needs to return an extra boolean as to whether a new zarr was actually created.

